PR #305, opened 2026-04-25
fix: separate TUI config from daemon config

### LINKED ISSUE #304: Bug: TUI overwrites daemon config by writing to ~/.clawvisor/config.yaml
## Summary
The TUI and daemon both write to `~/.clawvisor/config.yaml`, but they use incompatible YAML schemas. Running the TUI can overwrite the daemon/server config with TUI profile data.

## Expected
- Daemon/server config should live at `~/.clawvisor/config.yaml` and match the server schema documented by `config.example.yaml`.
- TUI client state should be stored separately in its own file.

## Actual
The TUI writes a file shaped like:
```yaml
profiles:
  default:
    server_url: http://localhost:25297
    token: ...
active_profile: default
poll_interval: 5
```
into `~/.clawvisor/config.yaml`, which clobbers the daemon config.

## Root cause
There are multiple config types in the repo, but two of them share the same default path:
- daemon/server config: `~/.clawvisor/config.yaml`
- TUI config: also `~/.clawvisor/config.yaml`

Relevant code paths:
- `internal/daemon/run.go` and `internal/daemon/setup.go` expect/write daemon config at `~/.clawvisor/config.yaml`
- `internal/tui/config/config.go` also points `DefaultPath()` at `~/.clawvisor/config.yaml`
- `cmd/clawvisor/cmd_tui.go` persists the refresh token there

## Proposed fix
Move TUI config to a different filename, for example `~/.clawvisor/tui.yaml`, so it cannot overwrite daemon config.

I have a small PR prepared for this approach.

jhferris:
Closes #304

## Summary
Move the TUI client config off `~/.clawvisor/config.yaml` and onto `~/.clawvisor/tui.yaml`.

## Why
The daemon/server config and the TUI config currently use incompatible schemas but share the same path. Running the TUI can overwrite daemon config with TUI profile data.

## Changes
- change TUI default config path to `~/.clawvisor/tui.yaml`
- update TUI refresh-token persistence comment
- update setup docs to reference the new TUI config file

## Notes
I did not add migration logic in this patch. In the current broken state, running the TUI already rewrites the old path, so separating the files is the urgent fix with the smallest diff.

## Verification
- `go test ./internal/tui/config ./cmd/clawvisor/...`

### DIFF
--- a/cmd/clawvisor/cmd_tui.go
+++ b/cmd/clawvisor/cmd_tui.go
@@ -149,7 +149,7 @@ func readLocalSession() (*localSessionFile, error) {
 	return &sess, nil
 }
 
-// saveRefreshToken persists the refresh token to ~/.clawvisor/config.yaml
+// saveRefreshToken persists the refresh token to ~/.clawvisor/tui.yaml
 // so subsequent TUI launches can use it directly.
 func saveRefreshToken(serverURL, refreshToken string) {
 	cfgPath, err := tuiconfig.DefaultPath()
--- a/docs/SETUP_LOCAL.md
+++ b/docs/SETUP_LOCAL.md
@@ -103,7 +103,7 @@ This covers:
 - **iMessage** — macOS only, reads the local `chat.db`
 
 The wizard generates `config.yaml` in the repository root and writes
-`~/.clawvisor/config.yaml` with the server URL so the TUI knows where to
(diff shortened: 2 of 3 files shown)


ericlevine:
Thank you for your PR! I'll get this released soon.

PR #306, opened 2026-04-25
feat(setup): prompt for bind host and port

jhferris:
## Summary
Add bind host and port prompts to the daemon setup flow so users can customize where Clawvisor listens during setup instead of being locked to `127.0.0.1:25297`.

## Changes
- prompt for bind host during interactive daemon setup
- prompt for bind port during interactive daemon setup
- write the selected host/port into generated `config.yaml`
- honor `SERVER_HOST` and `PORT` in non-interactive setup mode
- add test coverage for custom host/port config generation

## Why
Right now setup hardcodes `127.0.0.1:25297`, which is awkward for testing, alternate local ports, or installs that intentionally bind to another interface.

## Verification
- `go test ./internal/daemon/... ./cmd/clawvisor/...`


ericlevine:
Thanks for the PR! Is the thinking that adjusting these fields via the config.yaml is too opaque for new users who might not realize that the config is there?

ericlevine:
Closing this for now. Will revisit this in the future.

PR #309, opened 2026-04-29
fix(tui): render gateway log timestamps in local time

jhferris:
## Summary
Render TUI gateway log timestamps in the local system timezone instead of UTC.

## Root cause
Audit entries are stored in UTC, but the TUI formatted them directly without converting to local time first.

## Changes
- convert the gateway log list timestamp to `Local()` before formatting
- convert the gateway log detail timestamp to `Local()` before formatting


ericlevine:
Thank you for the submission!

PR #328, opened 2026-05-05
feat(runtime): per-session overrides for InlineApproval, lease TTL, harness allowlist

ericlevine:
## Summary

Extends `sessionRuntimeSettings` with three new optional fields realized from session metadata, each falling back to `cfg.RuntimePolicy.X` when unset:

- `InlineApprovalEnabled *bool` — gates inline-approval response parsing + `WouldPromptInline` event metadata
- `ToolLeaseTimeoutSeconds int` — per-session lease TTL override
- `HarnessAllowlist []string` — per-session host allowlist (replaces, doesn't augment, the cfg list)

## Why

The current proxy reads these values from `*config.Config` only — they're effectively global. Multi-tenant deployments (cloud-hosted runtime proxy) need per-org policy realized into per-session metadata at session-create time so a single binary serves users with different inline-approval, lease, and allowlist policies without a config reload.

Daemon mode is unaffected: when the metadata fields are absent (the common case for daemon-issued sessions), the accessors return the global cfg value, exactly as today.

## Changes

- `sessionRuntimeSettings` gains three new fields, all `omitempty`. Wire-compatible with existing serialized sessions — old metadata blobs parse cleanly, new fields default to "use cfg".
- New accessors: `sessionInlineApprovalEnabled`, `sessionToolLeaseTTL`, `sessionHarnessAllowlist`. Each takes `(*store.RuntimeSession, *config.Config)` and returns the per-session override when set, cfg fallback otherwise.
- Switched call sites in `tooluse_runtime.go` (6 InlineApproval reads, 2 lease TTL reads).
- `policy_hook.go`'s `isHarnessAllowlisted` now delegates to a new session-aware variant (`isHarnessAllowlistedForSession`); built-in defaults (anthropic.com/openai.com/chatgpt.com) and `cfg.LLM.Endpoint` still pass regardless.
- `session_settings_test.go` covers each accessor + override-vs-fallback semantics.

## Open semantic question

`HarnessAllowlist` per-session override **replaces** rather than augments the global. Rationale: enterprise customers set tight allowlists and need the override to be authoritative. Hobbyists who set their own allowlist on top of defaults can include the defaults explicitly. Open to feedback if "augment" is the better semantic.

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `go test ./...` — all packages pass, including the new `session_settings_test.go`
- [ ] CI green

### DIFF
--- a/pkg/runtime/proxy/policy_hook.go
+++ b/pkg/runtime/proxy/policy_hook.go
@@ -70,7 +70,7 @@ func (s *Server) InstallEgressPolicy(hooks PolicyHooks) {
 		req.ContentLength = int64(len(bodyBytes))
 
 		host := requestHost(req)
-		if isHarnessAllowlisted(hooks.Config, host) {
+		if isHarnessAllowlistedForSession(st.Session, hooks.Config, host) {
 			return req, nil
 		}
 		reqFingerprint := fingerprintRequest(st.Session.AgentID, req, bodyBytes)
@@ -730,6 +730,14 @@ func firstNonEmpty(values ...string) string {
 }
 
 func isHarnessAllowlisted(cfg *config.Config, host string) bool {
+	return isHarnessAllowlistedForSession(nil, cfg, host)
+}
+
+// isHarnessAllowlistedForSession returns whether the given host is on the
+// harness allowlist for this session. Per-session allowlist override (when
+// non-empty) takes precedence over cfg.RuntimePolicy.HarnessAllowlist; the
+// LLM endpoint and the built-in defaults always pass through regardless.
+func isHarnessAllowlistedForSession(session *store.RuntimeSession, cfg *config.Config, host string) bool {
 	host = strings.ToLower(strings.TrimSpace(host))
 	if host == "" {
 		return false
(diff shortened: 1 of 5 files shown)


ericlevine:
Force-pushed `45dde3c` addressing both findings.

**Finding 1 — empty `harness_allowlist` override is now expressible.**
Changed `HarnessAllowlist []string` → `*[]string`. JSON round-trips preserve the nil-vs-empty distinction (absent → nil pointer, `[]` → non-nil pointer to empty slice, `["x"]` → populated). Both the parse path and `sessionHarnessAllowlist` now gate on `!= nil` instead of `len(...) > 0`. New test cases cover:
- Explicit empty in metadata yields a non-nil empty slice from the accessor (not cfg fallback).
- `isHarnessAllowlistedForSession` rejects cfg-allowlist hosts under explicit empty override; built-in defaults (anthropic.com etc.) still pass.

**Finding 2 — held-approval surface + prompt are session-aware.**
`approvalSurface`, `renderHeldToolUsePrompt`, `renderExistingHeldPrompt` now take `*store.RuntimeSession` and route through `sessionInlineApprovalEnabled`. Three callsites updated. New `TestApprovalSurface_RespectsSessionOverride` exercises both directions:
- session=inline / cfg=dashboard → "inline" + inline prompt
- session=dashboard / cfg=inline → "dashboard" + dashboard prompt

Build + tests green. Force-push because the previous push accidentally swept some unrelated files (gitignored in the parent repo but visible to the submodule's git — fixed by adding `.context/` to the submodule's exclude list).

PR #344, opened 2026-05-06
feat: Microsoft 365 integration (Outlook, OneDrive) with full OAuth2 support

extremecoder-rgb:
# PR Description

This PR introduces comprehensive support for Microsoft 365 services via the Microsoft Graph API. It adds core infrastructure for Microsoft integrations and two fully-verified service adapters (Outlook, OneDrive), with initial support for Teams included but disabled for further verification.

### Verified Service Adapters
*   **Outlook**: Full support for listing/getting events and sending messages, featuring custom Go overrides for robust timezone and date formatting.
*   **OneDrive**: Support for file listing, metadata retrieval, and content upload/download. Includes a robust fix for cross-host 401 redirects and support for path-based downloads.

### Initial/Preview Adapters
*   **Teams**: Full adapter logic and unit tests are included, but the service is currently **disabled** in the runtime (via `.disabled` YAML and registration removal) until it can be verified against a live production tenant.

### Robust OAuth2 Infrastructure
*   **Unified Provider**: Implemented a secure Microsoft-specific OAuth2 provider with automated background token refresh.
*   **Identity Resolution**: Standardized account identity resolution using `userPrincipalName` across all services for a unified shared-credential model.
*   **Vault Integration**: Integrated with the system vault for secure credential storage.

### Safety & Refinement
*   **Panic Protection**: All Outlook overrides use safe type assertions to prevent nil-dereference crashes on missing or null API fields.
*   **Clean Branding**: Added official Microsoft Fluent SVG icons.
*   **Registry registration**: Services are properly registered via `goOverrides` in `pkg/clawvisor/defaults.go` for premium logic handling.

## Testing & Verification
*   **Live Verified**: Outlook and OneDrive have been verified end-to-end against real Microsoft 365 accounts.
*   **Unit Tests**: All adapters (`onedrive`, `outlook`, `teams`) include extensive `httptest` suites validating payload shapes and authentication logic.
*   **Loader Tests**: Verified that the YAML loader correctly handles the 15 active services.

---


ericlevine:
Thanks so much for putting this together, @extremecoder-rgb — this is a substantial contribution and we really appreciate the effort. Microsoft 365 support is something we've wanted to add, and the structure you've followed (hybrid YAML + Go overrides, shared credential helpers, per-service adapter packages) closely mirrors how our Google and Dropbox integrations are organized. That made the diff easy to read.

Before we can merge, there are a couple of larger issues we'd want addressed, plus some smaller polish items.

### Blockers

**1. Auth model needs OAuth2 with refresh tokens, not `api_key` Bearer.**

Microsoft Graph access tokens expire after ~1 hour, so the current `auth.type: api_key` setup means a user would paste a token from Graph Explorer and have it stop working before they finished their first task. For comparison, `internal/adapters/definitions/google_gmail.yaml` uses `auth.type: oauth2` with scopes, vault keys, conditional scopes, and refresh-token storage — and `pkg/clawvisor/defaults.go:161-170` shows how the Gmail/Drive/Contacts adapters receive an `oauthProvider` so they can transparently refresh. We'd want Microsoft to follow the same shape (auth-code flow against `https://login.microsoftonline.com/common/oauth2/v2.0/authorize` is probably the closest analogue to what we already do for Google).

This is the main thing — without it, the integration isn't usable in practice.

**2. The SVG icons need to be official Microsoft brand assets.**

Our existing logos in `web/public/logos/` are full-color brand-accurate SVGs (see `google-gmail.svg`, `slack.svg`, `notion.svg`). The three SVGs in this PR look like generic Lucide icons (stroke-only, `currentColor`, no brand palette) — a cloud, envelope, and people icon. They'd look noticeably out of place next to the other logos. Could you swap these for the official Outlook / OneDrive / Teams Fluent SVGs? Microsoft publishes brand assets at https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks.

### Smaller items

**3. `truncate()` is duplicated.** The helper you added in `internal/adapters/microsoft/credential.go` already exists in `internal/adapters/google/gmail/adapter.go`, `internal/adapters/dropbox/adapter.go`, and a few others. Could you lift it into `internal/adapters/format/` and have the existing call sites use it too? Happy to do that cleanup in a follow-up if you'd rather keep this PR focused.

**4. Tests are a bit thin.** The current tests cover only the "invalid token" and "unsupported action" paths. Consistent with our other adapter tests, so not strictly required to expand — but given this is a brand-new integration with no production usage yet, some httptest-server coverage of the Graph payload shapes (especially the `sendMail` envelope and the OneDrive upload PUT) would give us more confidence on landing.

### What's working well

- YAML + Go split is correct (YAML for plain GETs, Go for MIME/binary/nested-payload actions).
- Registration in `pkg/clawvisor/defaults.go` is wired up properly.
- YAML schema usage (`identity`, `verification_hints`, `risk`, `params.location`) matches our existing definitions.
- Risk categories and sensitivity levels look right for each action.

Once the OAuth refactor and the icons are sorted we should be in good shape to merge. Let us know if you'd like guidance on any of the above — happy to point at specific spots in the Google adapter as a reference if that helps.

extremecoder-rgb:
@ericlevine Thanks for the thorough review and guidance! I completely agree with the feedback. I've just pushed a few commits that address all the blockers and polish items

ericlevine:
Thanks for the quick turnaround on this, @extremecoder-rgb — the OAuth2 refactor looks great, the icons are now properly branded, and lifting `truncate` into `format/` (with the rune-aware implementation) is a nice cleanup beyond what we asked for. We're close.

A few things still need attention before we can merge. Apologies I didn't catch these the first time around.

### Blocker

**Tests don't compile.** `internal/adapters/microsoft/credential_test.go:42` still references the old `ExtractToken` function, which the refactor replaced with `Parse` / `Validate` / `HTTPClient`. Locally:

```
vet: internal/adapters/microsoft/credential_test.go:42:16: undefined: ExtractToken
FAIL    github.com/clawvisor/clawvisor/internal/adapters/microsoft [build failed]
```

The three sub-package tests (`onedrive`, `outlook`, `teams`) all pass — just this one file needs to be rewritten against the new API or removed.

### Account identity is inconsistent across the three services

All three Microsoft adapters share the `microsoft` vault key (good — that's the intended `scope_merge` design), but they disagree on how to identify the account:

- `microsoft_outlook.yaml:9` → `field: "mail"`
- `microsoft_onedrive.yaml:9` → `field: "mail"`
- `microsoft_teams.yaml:9` → `field: "id"`

The vault key is built from the alias (`pkg/adapters/adapters.go:418`), so the same Microsoft account would land under `microsoft:user@company.com` for Outlook/OneDrive and `microsoft:<aad-guid>` for Teams — splitting one account into two vault entries and breaking the shared-credential model.

There's also a related issue: `mail` can be null on personal MSA accounts and some B2C/workplace configs. When that happens, the generic identity resolver (`pkg/adapters/yamlruntime/adapter.go:196`) falls through and the handler keeps the alias as `default` (`internal/api/handlers/services.go:1233`), which can collapse multiple Microsoft accounts onto one vault entry.

Your `credential.go:111` already does the right thing — falls back from `mail` to `userPrincipalName`. I'd suggest standardizing all three YAMLs on `userPrincipalName` (always present, human-readable) so the YAML-driven identity flow has the same safety as the Go helper.

### OData `$` prefix is missing on three list actions

Microsoft Graph requires `$top`, `$filter`, `$search`, `$select`, `$skip` for OData query parameters. You correctly added `map_to: "$top"` etc. on `microsoft_outlook.yaml` `list_events` and `microsoft_onedrive.yaml` `list_files` — but a few siblings were missed:

- `microsoft_outlook.yaml` `list_messages` (lines 47, 51, 55, 59, 64) — all five query params (`top`, `filter`, `search`, `select`, `skip`) need `map_to: "$..."`.
- `microsoft_teams.yaml` `list_messages` (line 90) — `top` needs `map_to: "$top"`.
- `microsoft_onedrive.yaml` `search_files` (line 119) — `top` needs `map_to: "$top"`.

Without these, Graph silently ignores the params, so paging and filtering won't work on those endpoints.

### Web UI doesn't expose Microsoft OAuth config

The backend ships the `/api/system/microsoft-oauth` endpoints (`internal/api/server.go:743`, `internal/api/handlers/services.go:2182`) and the daemon flow requires app credentials to activate Microsoft services (`internal/daemon/services_setup.go:379`) — but the dashboard has no UI for it:

- `web/src/api/client.ts` only has `getGoogleOAuth` / `setGoogleOAuth` (around line 1292) — no Microsoft equivalents.
- `web/src/pages/Settings.tsx:340` renders only a Google OAuth card.
- `web/src/pages/Services.tsx:1347` queries / `:1442` warns only about missing Google OAuth.

Net effect: dashboard users can attempt to connect Outlook/OneDrive/Teams without any pre-flight indication that Microsoft OAuth setup is missing, and there's no place in the UI to configure it. We could you mirror the Google patterns — API client methods, a Settings card, and a missing-status banner on the Services page.

If you'd rather do this as a follow-up (or defer it to me or another contributor), that's totally fine too.

### Smaller items

- The PR description still says the implementation "uses api_key (Bearer token) authentication" — the actual code is OAuth2 with app credentials. Could you update the description so reviewers and future readers aren't misled?
- `HTTPClient` doesn't call `Validate(credBytes)` before constructing the OAuth2 client, so invalid creds only fail when the first Graph request goes out (the `TestExecute_InvalidToken` tests are passing because the OAuth library hits `login.microsoftonline.com` and that fails). Calling `Validate` up-front would make these tests fail fast and not depend on network.

### What's good

To be clear, a lot of this PR is in great shape:

- OAuth2 wiring (`Stored` struct, refresh-token expiry, `oauth2.NewClient` auto-refresh, vault provider with env-var fallback, system-vault key, `/api/system/microsoft-oauth` endpoints, TUI prompt flow) all mirror the Google pattern correctly.
- `endpoint: microsoft` correctly falls through to the schema's default branch and uses the YAML's `authorize_url` / `token_url` (`pkg/adapters/yamlruntime/adapter.go:247-255`).
- The httptest-based payload coverage (`TestSendMessage`, `TestCreateEvent`, `TestUploadFile`, etc.) is exactly what we wanted.
- Logos are real Microsoft brand assets now.
- Per-action OAuth scopes are wired up cleanly.

Once the build, identity alignment, OData prefixes, and dashboard UI are sorted, I think we're in good shape. Let me know if you'd like guidance on any specific piece — happy to point at the Google equivalents as references.

Thank you so much for your contribution! I'm excited to land this.

ericlevine:
@extremecoder-rgb Let me know if you want me to take over the remaining items and ship this. But would also love to support you landing this.

extremecoder-rgb:
> @extremecoder-rgb Let me know if you want me to take over the remaining items and ship this. But would also love to support you landing this.

I will like to work on it sir If I get stuck I will take your help surely

extremecoder-rgb:
@ericlevine I have done relevant fixes. Let me know if everything is correctly implemented or not

ericlevine:
Thanks @extremecoder-rgb! The latest commit cleared every code-review item, and I took the branch live against a real Microsoft account to exercise it end to end. The OAuth flow, Settings card, scope refresh, and identity aliasing all work as intended. A few real bugs surfaced under live testing, plus two small build-time fixes I had to apply locally to even run it. None of these are huge, but they should be addressed before we merge.

### Local fixes I had to apply to test the branch

Before I could run the daemon and the dashboard, two small things were broken:

**1. `pkg/adapters/yamlloader/loader_test.go` — `TestLoadEmbeddedDefinitions` fails.** The test enumerates the full set of expected adapters and would either fail (extras present) or silently miss the new ones. The three Microsoft adapters need to be added:

```go
"microsoft.onedrive": false,
"microsoft.outlook":  false,
"microsoft.teams":    false,
```

**2. `web/src/pages/Services.tsx` — TypeScript build error in `AddServiceModal`.** You added `microsoftOAuthMissing: boolean` to the prop *type*, and you reference `microsoftOAuthMissing` in the JSX at the "Configure in Settings" branch — but the destructured prop list at the top of the function is missing it, so the symbol is undeclared:

```tsx
function AddServiceModal({
  services,
  onClose,
  onSuccess,
  googleOAuthMissing,
  microsoftOAuthMissing,  // ← needs to be added
  activeConnectionCount,
}: { ... })
```

Worth running `go test ./...` and the web `tsc`/build to catch any other latent issues before the next push.

### Real adapter bugs from live testing

**1. OneDrive `download_file` returns 401 on every file** (adapter-wide). Tested against a freshly uploaded sandbox file *and* an unrelated pre-existing PDF — both fail with `ADAPTER_ERROR: status 401`, while `list_files` and `get_file` succeed on the same account/task.

This is the well-known Graph + Go OAuth interaction ([golang/oauth2#308](https://github.com/golang/oauth2/issues/308)): `/me/drive/items/{id}/content` 302-redirects to a pre-signed SharePoint URL on a different host. Go's stdlib strips `Authorization` on cross-host redirects, but the `oauth2` `RoundTripper` re-injects it on the redirected request — SharePoint's download host rejects the Bearer token with 401.

The clean fix is to fetch metadata with `?$select=@microsoft.graph.downloadUrl` and download from that pre-signed URL using a plain (non-OAuth) HTTP client. Alternatively, set a `CheckRedirect` on the client that strips `Authorization` on host change.

**2. Outlook `create_event` silently drops `attendees`.** Test sent `["email@gmail.com"]` (JSON array) with a fully-formed event; `get_event` afterwards returned `attendees: []` with no error or warning surfaced.

In `outlook/adapter.go:createEvent`:

```go
attendeesStr, _ := params["attendees"].(string)
...
for _, email := range strings.Split(attendeesStr, ",") {
```

The chain-context layer passes the param as a JSON array, the `string` type-assertion fails silently, and `attendeesStr` becomes `""`. Need to handle both `string` (comma-separated) and `[]any` / `[]string`, or change the YAML to declare the param as an array type.

**3. Teams blocked at task creation: `MISSING_SCOPES: https://graph.microsoft.com/Team.ReadBasic.All`.** The YAML at `microsoft_teams.yaml:14-19` does request the scope, so the connected service shouldn't be missing it. Could you reproduce this locally and capture the actual scope set returned after activation? Possibilities:

- The OAuth flow built the consent URL without the full scope list (worth diffing how `auth.oauth.scopes` is passed against how Google's `conditional_scopes` are handled).
- A scope-merge bug where activating one Microsoft service overwrote another's stored scopes.
- Tenant-side admin-consent on Teams scopes (out-of-band, but should at least be documented in the setup doc).

Until we know which one, none of `list_teams`, `list_channels`, `list_messages`, `send_message` can be exercised.

### Smaller polish items (non-blocking, worth catching now)

- **`upload_file` rejects empty `content`.** Empty files are valid in OneDrive — the validation should allow them.
- **`get_file` doesn't return `path`, but `upload_file` does.** Pick one shape.
- **Invalid `item_id` returns the raw Graph `400 invalidRequest`.** A friendlier "not found" mapping would be nicer.
- **`get_event` returns tz-naive UTC strings** (e.g. `"2026-05-09T21:00:00.0000000"`) even when the user submitted `timezone: "America/Los_Angeles"`. Conversion is correct (14:00 PT → 21:00 UTC) but consumers shouldn't have to assume UTC. Either return tz-aware ISO timestamps, or echo `timezone` separately on read.
- **`create_event` response only echoes `{end, id, start, subject}`.** Adding `location/body/attendees/timezone` would let agents skip a `get_event` round-trip — the data is already persisted server-side.
- **No resumable-upload support.** `upload_file` only exercises Graph's simple upload, which caps at 4 MB. Either implement upload sessions for larger files, or return a clear "file too large" error before hitting Graph.

### Out of scope (mentioning so you know they're tracked elsewhere)

A few issues showed up in testing that are platform-level, not yours to fix in this PR:

- Chain-context not propagating `id` from `create_event` / `upload_file` responses, forcing a `create → list → get` pattern instead of `create → get`. Same class as the iMessage `get_thread` issue.
- Adapter validation errors return `code: null` instead of the documented machine-readable codes.

I'll file those separately.

### Recommendation

Fix items 1, 2, and the two local-build issues, plus get item 3 (Teams scopes) reproduced and either fixed or clearly documented. If you can, I'd strongly encourage you to test this locally as well to make sure it's working end-to-end. After that I think we're ready to merge.

extremecoder-rgb:
@ericlevine Thanks for the incredibly detailed review! I've just pushed a comprehensive set of fixes addressing all the items you found during live testing:

1. Local Build Fixes:

Added the three Microsoft adapters to the loader_test.go suite.
Fixed the microsoftOAuthMissing prop destructuring in Services.tsx to resolve the TypeScript error.
2. Real-World Bug Fixes:

OneDrive Download (401 Fix): Refactored download_file to fetch the pre-signed @microsoft.graph.downloadUrl and use a plain HTTP client. This successfully bypasses the oauth2.RoundTripper redirect conflict.
Outlook Attendees: Updated create_event to handle both strings and JSON arrays. I've also hardened the type-assertion logic to prevent any silent failures.
Teams Scopes: Switched all YAML scope definitions to "short names" (e.g., Team.ReadBasic.All). This should resolve the MISSING_SCOPES match failure you encountered.
3. Polishing & Robustness:

OneDrive: Allowed empty file uploads, added a 4MB size limit check for simple uploads, and mapped invalid item_id errors to a friendly not_found.
Outlook: Enhanced the create_event response to echo all metadata (location, body, attendees, etc.) and added the timezone field to the get_event output.
Refinements: Based on bot feedback, I've also refined the error mapping and parameter validation to be more specific and safe.
I've run the full Go test suite locally (pkg/adapters/yamlloader and internal/adapters/microsoft/...) and every component passed 100%. Ready for the final merge!

ericlevine:
@extremecoder-rgb Great! Thank you so much! Have you had a chance to test it locally end-to-end with a real Microsoft OAuth? If not, I can do it, but it would be great for you to give it a try if you haven't.

extremecoder-rgb:
@ericlevine I've run the full local Go test suite (pkg/adapters/yamlloader, internal/adapters/microsoft/...) and verified that every fix—including the new pre-signed URL download logic and the hardened attendees handling—is passing 100%.

I hit some delays getting my personal Azure sandbox qualified for a live test today, but I'm confident the logic is solid. Since you already have your live test account connected, could you give it one final spin? It should be rock solid now!

ericlevine:
@extremecoder-rgb OK! So I was able to test Outlook and OneDrive, and both are mostly working. The download file functionality is not working for OneDrive, and I'm not able to test Teams at all with my test account. We could either remove Teams and the OneDrive download functionality and ship as is, or once you get your test setup working you can address the remaining OneDrive download bug, test with Teams, and we can ship the whole thing.

extremecoder-rgb:
Screenshot 2026-05-09 074656
Hi @ericlevine ,

I have refactored the code to address the issues you identified:

OneDrive Download: Fixed the 401 redirect issue by stripping the Authorization header on cross-host requests. It now also supports path-based downloads.
Teams: I've adjusted the scopes to be as minimal as possible (Team.ReadBasic.All) to satisfy the security checks.
Outlook: Added Go overrides for better date and event formatting.
Note on Testing: All Go unit tests are passing. However, I haven't been able to test end-to-end because I don't currently qualify for a Microsoft 365 Developer sandbox (as shown in my previous screenshot).

Next Step: Could you please test this one last time? If OneDrive or Teams are still not working for you, let me know—I am happy to remove those specific components from this PR so we can merge the core Outlook and identity integration first. I can then revisit the remaining bugs in a separate PR once I have a working test setup.

Thanks!

ericlevine:
Hi @extremecoder-rgb, I can confirm that the download file action is working as expected. Great work!

I didn't create a Microsoft developer account to test, I created a free Azure account and went to https://entra.microsoft.com/. Do you have an account that uses Microsoft teams?

extremecoder-rgb:
@ericlevine No, I don't have a Microsoft Teams account to test with — I only have the free Azure/Entra ID account

ericlevine:
Thanks @extremecoder-rgb — the OneDrive `download_file` fix is exactly right (pre-signed `@microsoft.graph.downloadUrl` for the primary path, `CheckRedirect` stripping `Authorization` as the fallback), and echoing `timezone` on the new `list_events` / `get_event` overrides resolves the tz-naive issue from the live test. Outlook and OneDrive are both working end-to-end against a real account now.

A few items before this is ready to merge.

### Teams: disabling for this PR

I haven't been able to live-test Teams against a real Microsoft 365 tenant, and from our discussion neither have you. The two existing live-test passes both surfaced real bugs that survived code review (Outlook's `attendees` drop, OneDrive's 401 redirect), so I'm not comfortable shipping a third Graph adapter that no one has exercised against the real API — especially given the specific risks I can spot by inspection:

- `list_messages` field paths like `from.user.displayName` will likely blow up on bot/system messages where `from.user` is null.
- `list_channels` uses `/teams/{id}/channels`, which only returns standard channels — private and shared channels are silently filtered out.
- `send_message` always sends `contentType: html` with no encoding of HTML-special characters in the payload.
- And the scope-grant flow that bit us twice on Outlook/OneDrive setup will bite the first Teams operator just as hard, without docs.

I'd like to ship the rest of this PR with Teams effectively disabled until someone can run the four Teams actions against a real tenant. The cleanest way is to take the YAML out of the embed glob:

1. `git mv internal/adapters/definitions/microsoft_teams.yaml internal/adapters/definitions/microsoft_teams.yaml.disabled`
2. Remove the 2-line `teams := teamsadapter.New(...)` block and the `teamsadapter` import from `pkg/clawvisor/defaults.go`.
3. Remove `"microsoft.teams": false,` from `pkg/adapters/yamlloader/loader_test.go:37`.
4. Remove the `microsoft.teams` row from the README services table.
5. Leave `internal/adapters/microsoft/teams/*.go` and the SVG in place — the package still compiles and tests still run, but with no YAML loaded and no registration, the runtime has no way to reach it.

Re-enabling is the inverse: rename the file back, restore the registration, add the test entry, restore the README row. Five small lines.

Happy to push these as a follow-up commit on a branch I can submit against yours if that's easier than you doing it.

### Two new code issues

**1. Nil-deref risk in the new Outlook overrides.** `listEvents` and `getEvent` in `internal/adapters/microsoft/outlook/adapter.go` do unchecked type assertions on map values:

```go
"subject": format.SanitizeText(item["subject"].(string), format.MaxFieldLen),
"body":    format.SanitizeText(body["content"].(string), format.MaxBodyLen),
```

If Graph returns an event with a missing or null `subject` (uncommon but valid), `item["subject"]` is `nil` and `nil.(string)` panics at runtime. Same pattern for `body["content"]`. The previous YAML-driven path handled missing fields gracefully via the response renderer; the new Go path needs explicit guards. The comma-ok form everywhere fixes it:

```go
subject, _ := item["subject"].(string)
"subject": format.SanitizeText(subject, format.MaxFieldLen),
```

**2. Stale `verification_hints` in `microsoft_teams.yaml`.** Commit `55557f9f` rolled `Group.Read.All` out of the scope list, but the hints block still references it:

> *"IMPORTANT: Ensure your Azure App registration has 'Team.ReadBasic.All' and 'Group.Read.All' permissions added and consented."*

Drop the `Group.Read.All` mention so the docs match the actual requested scopes. (This still applies even though the YAML is being disabled — the hint is wrong if/when it comes back.)

### Net state

Once Teams is disabled and the two small items above are addressed, this is ready to merge. Outlook and OneDrive both work against a real account, the OAuth refactor is solid, the rebase on `main` is clean, and the operator-facing UI is wired up.

Once Teams gets verified, it's a small reverse-of-the-rename PR to re-enable. I'll file a tracking issue for that.

extremecoder-rgb:
Thanks @ericlevine, I really loved working on shipping this Microsoft integration for Clawvisor. It was a great learning experience, especially digging into the Graph API quirks like the OneDrive redirects and the Outlook formatting.

I’ve applied all your final suggestions—the Outlook safety fixes are in, and Teams is now cleanly parked as a 'preview' feature until we can get it verified live. I've also updated the PR title and description to reflect the final state.

Really appreciate your guidance and the live-testing help to get this over the finish line. Excited to see this merged!

ericlevine:
Thanks @extremecoder-rgb — verified the Outlook nil-deref fixes, the Teams disable, and the `verification_hints` cleanup all landed correctly. Rebased on `main` locally, build/vet/tests all green.

Two small items before we merge:

**1. README markdown table is broken.** The Teams row was replaced with an empty line at that row's position instead of being deleted entirely:

```
| `microsoft.onedrive` | OneDrive | ... |
                                              ← blank line here
| `github` | GitHub | ... |
```

GitHub treats a blank line inside a markdown table as a terminator, so the rows from `github` onward render as a separate, headerless table. Just delete the empty line so the rows are contiguous.

**2. We just merged a [`CONTRIBUTING.md`](https://github.com/clawvisor/clawvisor/blob/main/CONTRIBUTING.md) (#362) — worth a read.** It landed after you opened this PR, so it would be great for you to review it and acknowledge it before we merge your PR. The most important thing to know: it has an implicit inbound=outbound clause — *"By submitting a contribution, you agree that your contribution is licensed under the same terms [Elastic License 2.0] and that you have the right to submit it."*

Once the README line is gone, this is ready to merge. Thanks for sticking with it through all the rounds of feedback — really nice work.

extremecoder-rgb:
Thanks @ericlevine 

I've removed that blank line from the README, so the services table should now render as a single contiguous block.

I also acknowledge the new CONTRIBUTING.md and agree that this contribution is licensed under the Elastic License 2.0 terms.

Really appreciate you sticking with me through all the feedback—happy to see this land!

ericlevine:
@cubic-dev-ai review

ericlevine:
Thank you again, @extremecoder-rgb! Excited to support these services in Clawvisor.

extremecoder-rgb:
Thanks @ericlevine , Happy that I contributed something meaningful

PR #348, opened 2026-05-06
feat(api): per-user FeaturesHook for plan-based feature gating

ericlevine:
## Summary

Replace the frontend-only plan check from #347 with a server-side mechanism so any consumer of `features.local_daemon` (Settings, Services, etc.) gets the correct gate without duplicating billing logic in the UI.

- Add a `FeaturesHook` option on `pkg/clawvisor.ServerOptions` that lets a cloud/enterprise layer override the `FeatureSet` returned by `/api/features` on a per-user basis.
- Add a new `OptionalUser` middleware (`internal/api/middleware/auth.go`) that populates the request user when a valid JWT is present and falls through silently otherwise. Wire it onto `/api/features` so the bootstrap (pre-login) call still works.
- Frontend (`web/src/hooks/useAuth.tsx`) refetches features whenever the authenticated user changes, so the per-user response replaces the anonymous bootstrap value after login (and resets on logout).
- Revert `Settings.tsx` to a plain `features?.local_daemon` check — the server is now authoritative.

The cloud-side hook implementation (gating `LocalDaemon` on plan + `LOCAL_DAEMON_EMAILS` allowlist) lives in `clawvisor/cloud` and will land alongside this.

## Stacking

This PR is based on `ericlevine/gate-local-daemon-ui-pro` (#347). Once #347 merges, GitHub will retarget this PR to `main` and the only remaining commit will be the FeaturesHook addition (the Settings.tsx revert becomes a no-op against post-#347 main).

## Test plan

- [ ] `go test ./...` passes (verified locally).
- [ ] `tsc --noEmit` and `vite build` clean.
- [ ] Manual: free-tier cloud user sees the local daemon section disappear after login (was visible during anonymous bootstrap).
- [ ] Manual: pro / enterprise / grandfathered users still see the section.
- [ ] Manual: self-hosted (no `FeaturesHook` registered) is unchanged.
- [ ] Manual: logout returns the FeatureSet to deployment defaults.

### DIFF
--- a/internal/api/middleware/auth.go
+++ b/internal/api/middleware/auth.go
@@ -68,6 +68,37 @@ func RequireUser(jwtSvc auth.TokenService, st store.Store) func(http.Handler) ht
 	}
 }
 
+// OptionalUser is middleware that validates a user JWT if present and injects
+// the user into the request context. Unlike RequireUser it never rejects the
+// request: missing, invalid, or purpose-restricted tokens cause the handler
+// to run without a user in context. Use this for routes that must be reachable
+// pre-login but want to vary their response when a user is authenticated
+// (e.g. /api/features for per-user feature gating).
+func OptionalUser(jwtSvc auth.TokenService, st store.Store) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


ericlevine:
Closing and reopening with a clean PR body (the original was created when this was stacked on #347; rebase is done and #347 is already merged, so the stacking notes are stale).

PR #363, opened 2026-05-11
fix(mcp): url-encode service param to prevent query string injection

adityajha2005:
## Summary

Fixes a query string parameter injection vector in the `fetch_catalog` MCP tool route. Agent-controlled input was concatenated directly into a URL without encoding, allowing an agent to inject arbitrary query parameters into an internal route.

## Security Issue

The `fetch_catalog` tool accepts a `service` argument from the agent and builds an internal route URL:

```go
// before
path += "?service=" + svc
```

An agent could pass a value like `google.gmail&admin=true`, which the server would parse as two separate query parameters:

```
/api/skill/catalog?service=google.gmail&admin=true
```

The injected parameter (`admin=true`) would be invisible to the caller but visible to the internal route handler - a classic query string injection pattern.

### Why this matters

Clawvisor's core security model treats agents as untrusted. An agent that can silently inject parameters into internal routes breaks that boundary. The `fetch_catalog` handler may not act on unknown params today, but:

- Future additions to this route (pagination, filtering, access flags) would be immediately exploitable with no further changes on the agent side
- The injection point exists silently - there is no error, no log, no indication the parameter was injected
- It establishes a pattern of unencoded agent input reaching internal routing, which is wrong regardless of current impact

## Fix

Wrap the value with `url.QueryEscape` so special characters are encoded as literal data:

```go
// after
path += "?service=" + url.QueryEscape(svc)
```

`&` becomes `%26`, `=` becomes `%3D` - the handler receives a single `service` param with the full raw value, injection attempt and all. The route behaves correctly for all valid service identifiers since they contain no special characters.

### DIFF
--- a/internal/mcp/tools_exec.go
+++ b/internal/mcp/tools_exec.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 )
 
@@ -123,7 +124,7 @@ func buildInternalRequest(toolName string, arguments json.RawMessage) (internalR
 	case "fetch_catalog":
 		path := "/api/skill/catalog"
 		if svc := getString("service"); svc != "" {
-			path += "?service=" + svc
+			path += "?service=" + url.QueryEscape(svc)
 		}
 		return internalRoute{"GET", path, "GET /api/skill/catalog", nil}, nil, nil
 


adityajha2005:
Hey @ericlevine - small security fix here                                                                                 
                                                                                                                                
  The `fetch_catalog` MCP tool was concatenating the agent-supplied `service` value directly into the query string without encoding. An agent could pass something like `google.gmail&skip_verification=true` and the server would parse the injected param as a real input.                                                                                                               
                                                                                                                                  
  No known exploit today since the handler likely ignores unknown params, but given this is an open source project and agents are explicitly untrusted in Clawvisor's model, leaving a raw injection point felt wrong. Especially if new params get added to that route down the line.                                                                                                            
                                                                                                                                  
  Fix is a one-liner - `url.QueryEscape` on the value before it hits the URL. Happy to add a test case if that would help get this merged.

ericlevine:
Thanks for catching this, @adityajha2005 — the analysis in the PR body is spot-on, and the fix is the right call.

Before I merge, one thing worth flagging: `buildWaitQuery` (line ~220) has the same pattern with the `timeout` parameter:

```go
q += "&timeout=" + t
```

The tool schemas declare `timeout` as an integer, but since the threat model treats agents as untrusted, a malicious agent can send `"timeout": "120&admin=true"` as a JSON string and `getString` will return it verbatim. Same injection class, just via a different parameter, affects `create_task`, `get_task`, `expand_task`, `gateway_request`, `gateway_batch`, and `execute_request`.

A few options, your call:

1. Extend this PR to cover `timeout` as well, plus the test you offered (covering both `service` and `timeout` against `buildInternalRequest`).
2. Merge this as-is and you send a follow-up PR for the `timeout` fix.
3. Merge this as-is and I'll take the follow-up.

Happy with any of these, let me know which you'd prefer.

adityajha2005:
> Thanks for catching this, @adityajha2005 — the analysis in the PR body is spot-on, and the fix is the right call.
> 
> Before I merge, one thing worth flagging: `buildWaitQuery` (line ~220) has the same pattern with the `timeout` parameter:
> 
> ```go
> q += "&timeout=" + t
> ```
> 
> The tool schemas declare `timeout` as an integer, but since the threat model treats agents as untrusted, a malicious agent can send `"timeout": "120&admin=true"` as a JSON string and `getString` will return it verbatim. Same injection class, just via a different parameter, affects `create_task`, `get_task`, `expand_task`, `gateway_request`, `gateway_batch`, and `execute_request`.
> 
> A few options, your call:
> 
> 1. Extend this PR to cover `timeout` as well, plus the test you offered (covering both `service` and `timeout` against `buildInternalRequest`).
> 2. Merge this as-is and you send a follow-up PR for the `timeout` fix.
> 3. Merge this as-is and I'll take the follow-up.
> 
> Happy with any of these, let me know which you'd prefer.

sounds good to me! let's merge this and i'll send a follow-up PR covering the timeout param and the test cases. thanks for the review!

ericlevine:
Great! Thanks for the contribution and thank you in advance for the follow-ups. Merging now.

PR #366, opened 2026-05-11
fix(mcp): url-encode timeout param and add injection tests

adityajha2005:
## Summary

Follow-up to #[prev-pr-number] which fixed the `service` param in `fetch_catalog`. As flagged by @[maintainer] in review, `buildWaitQuery` had the same injection class via the `timeout` parameter. This PR fixes that and adds tests covering both params.

## The Bug

In `buildWaitQuery`, the agent-supplied `timeout` value was concatenated raw into the query string:

```go
// before
q += "&timeout=" + t
```

The tool schema declares `timeout` as an integer, but `getString` reads it as a string. An agent could send `"timeout": "120&admin=true"` as a JSON string and the URL becomes:

```
?wait=true&timeout=120&admin=true
```

The server sees `admin=true` as a real separate parameter. Same injection class as the `service` fix, just via a different parameter. Affects `create_task`, `get_task`, `expand_task`, `gateway_request`, `gateway_batch`, and `execute_request` - all tools that call `buildWaitQuery`.

## Fix

```go
// after
q += "&timeout=" + url.QueryEscape(t)
```

`&` becomes `%26`, `=` becomes `%3D`. The injected content stays inside the `timeout` value and never reaches the server as a separate param.

## Tests

Added `internal/mcp/tools_exec_test.go` covering both `service` and `timeout` params:

**service param cases:**
- normal value passes through unchanged
- `&` injection encoded correctly
- `=` injection encoded correctly
- spaces encoded correctly
- empty value omits the query string entirely

**timeout param cases:**
- normal value passes through unchanged
- `&` injection encoded correctly
- `=` injection encoded correctly
- no timeout omits the param entirely
- `wait=false` ignores timeout altogether
image

## Files Changed

| File | Change |
|------|--------|
| `internal/mcp/tools_exec.go` | Add `net/url` import, encode `timeout` param |
| `internal/mcp/tools_exec_test.go` | New file - injection tests for `service` and `timeout` |

### DIFF
--- a/internal/mcp/tools_exec.go
+++ b/internal/mcp/tools_exec.go
@@ -218,7 +218,7 @@ func buildWaitQuery(args map[string]json.RawMessage, getString func(string) stri
 	}
 	q := "?wait=true"
 	if t := getString("timeout"); t != "" {
-		q += "&timeout=" + t
+		q += "&timeout=" + url.QueryEscape(t)
 	}
 	return q
 }
--- a/internal/mcp/tools_exec_test.go
+++ b/internal/mcp/tools_exec_test.go
@@ -1,0 +1,134 @@
+package mcp
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func args(t *testing.T, kv map[string]any) json.RawMessage {
+	t.Helper()
+	b, err := json.Marshal(kv)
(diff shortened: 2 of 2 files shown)


ericlevine:
Thank you for the contribution!

PR #377, opened 2026-05-12
fix(intent): enhance tag stripping to handle case variations

adityajha2005:
Follow-up to #375 as suggested by @ericlevine.

## Summary

The previous PR stripped role tags like `<system>` and `<assistant>` from agent-supplied fields, but the matching was case-sensitive. An agent could bypass stripping by using `<SYSTEM>`, `<System>`, or any other casing variant.

## Fix

The `stripTags` function now searches against a lowercased copy of the input while cutting from the original, so casing of the actual content is preserved but tags are matched case-insensitively.

```go
// before - missed <SYSTEM>, <System>, etc.
s = strings.ReplaceAll(s, tag, "")

// after - matches any casing
lower := strings.ToLower(s)
// search in lower, cut from both s and lower at the same position
```

## Tests

Added 3 new cases to `TestBuildVerificationUserMessage_TagStripping`:

- `<SYSTEM>` (all caps) stripped
- `<System>` (mixed case) stripped  
- `<ASSISTANT>` (all caps) stripped

image

## Files Changed

| File | Change |
|------|--------|
| `internal/intent/prompts.go` | Case-insensitive tag matching in `stripTags` |
| `internal/intent/verifier_test.go` | 3 new uppercase/mixed-case test cases |

### DIFF
--- a/internal/intent/prompts.go
+++ b/internal/intent/prompts.go
@@ -277,8 +277,17 @@ The user has explicitly opted this task into lenient verification. Apply these a
 func buildVerificationUserMessage(req VerifyRequest) string {
 	params, _ := json.MarshalIndent(req.Params, "", "  ")
 
+	asciiLower := func(s string) string {
+		b := []byte(s)
+		for i, c := range b {
+			if 'A' <= c && c <= 'Z' {
+				b[i] = c + ('a' - 'A')
+			}
+		}
+		return string(b)
+	}
 	stripTags := func(s string) string {
-		lower := strings.ToLower(s)
+		lower := asciiLower(s)
 		for _, tag := range []string{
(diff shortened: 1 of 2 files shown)


ericlevine:
Looks great! One comment -- you could add a test for `<sy<system>stem>` to prove that it handles that case as well.

adityajha2005:
> Looks great! One comment -- you could add a test for `<sy<system>stem>` to prove that it handles that case as well.

added the `<sy<system>stem>` test as suggested, and also added a `<sy<SYSTEM>stem>` case to cover nested + uppercase reassembly together. let me know if anything else needed!

ericlevine:
Looks great! Merging now.

PR #387, opened 2026-05-15
docs: fix setup guide drift vs current codebase

adityajha2005:
Fixes the drift in `docs/SETUP_LOCAL.md` and `docs/SETUP_DOCKER.md` that the maintainer flagged — aligning the guides with how the codebase actually works today.

### `docs/SETUP_LOCAL.md`

**1. Wrong config.yaml location**
The guide said the setup wizard generates `config.yaml` in the repository root. It actually writes to `~/.clawvisor/config.yaml` — verified via `internal/daemon/run.go:ensureDataDir()`.

**2. Setup wizard coverage was outdated**
The guide listed Google OAuth and iMessage as interactive wizard steps. Neither is true anymore:
- Google OAuth (`GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`) is configured separately after setup — the wizard never prompts for it
- iMessage is auto-enabled on macOS and disabled elsewhere — no user input needed
- The wizard now covers: LLM provider, relay, and telemetry

**3. `PUBLIC_URL` description was incomplete**
Listed as "Used in Telegram notification links" only. Since #380, `PUBLIC_URL` also controls the magic link URL shown at startup and OAuth callback redirects.

**4. `SQLITE_PATH` env var missing**
`SQLITE_PATH` is a real env var handled in `pkg/config/config.go:441` but was not in the reference table.

### `docs/SETUP_DOCKER.md`

**5. `make docker-setup` + `make docker` undocumented**
The Makefile has a simpler single-container SQLite path (`docker-compose.local.yml`) using `make docker-setup` and `make docker`. This was completely absent from the Docker guide — users who looked at the Makefile had no docs to match it.

---

## Changes

| File | Change |
|------|--------|
| `docs/SETUP_LOCAL.md` | Fix config.yaml path (repo root → `~/.clawvisor/`) |
| `docs/SETUP_LOCAL.md` | Remove Google OAuth from wizard steps; add relay/telemetry |
| `docs/SETUP_LOCAL.md` | Clarify iMessage is auto-detected, not wizard-prompted |
| `docs/SETUP_LOCAL.md` | Update `PUBLIC_URL` description |
| `docs/SETUP_LOCAL.md` | Add `SQLITE_PATH` to env vars table |
| `docs/SETUP_DOCKER.md` | Document `make docker-setup` + `make docker` as the simpler SQLite path |

---

No code changes — docs only.


ericlevine:
Thanks for cleaning this up. The direction looks right, and the docs are definitely closer to the current setup flow now. I found a few places where the new text still seems to drift from the actual commands users will run:

1. In `docs/SETUP_LOCAL.md`, the wizard path update to `~/.clawvisor/config.yaml` looks correct, but Step 4 still tells users to run `make run`. That target runs `bin/clawvisor server`, which falls back to `config.yaml` in the repo unless `CONFIG_FILE` is set, so it won’t automatically use the config generated by `make setup`. I think this should either set `CONFIG_FILE="$HOME/.clawvisor/config.yaml"` in the command or use the daemon entrypoint that already points at that config.

2. The new relay wording says it is “opt-in”, but the current setup code initializes relay as enabled by default and does not prompt interactive users before registering/writing `relay.enabled: true`. Could you adjust the wording to say relay is enabled by default and can be disabled in config?

3. In `docs/SETUP_DOCKER.md`, the new SQLite single-container path says to skip to Step 5 for the magic link, but Step 5 uses `docker compose -f deploy/docker-compose.yml logs app`. `make docker` runs `deploy/docker-compose.local.yml`, so that log command targets the wrong compose file for the SQLite path. Maybe add a separate magic-link extraction command for the local compose file, or tell users to copy it from the foreground `make docker` output.

adityajha2005:
> Thanks for cleaning this up. The direction looks right, and the docs are definitely closer to the current setup flow now. I found a few places where the new text still seems to drift from the actual commands users will run:
> 
> 1. In `docs/SETUP_LOCAL.md`, the wizard path update to `~/.clawvisor/config.yaml` looks correct, but Step 4 still tells users to run `make run`. That target runs `bin/clawvisor server`, which falls back to `config.yaml` in the repo unless `CONFIG_FILE` is set, so it won’t automatically use the config generated by `make setup`. I think this should either set `CONFIG_FILE="$HOME/.clawvisor/config.yaml"` in the command or use the daemon entrypoint that already points at that config.
> 2. The new relay wording says it is “opt-in”, but the current setup code initializes relay as enabled by default and does not prompt interactive users before registering/writing `relay.enabled: true`. Could you adjust the wording to say relay is enabled by default and can be disabled in config?
> 3. In `docs/SETUP_DOCKER.md`, the new SQLite single-container path says to skip to Step 5 for the magic link, but Step 5 uses `docker compose -f deploy/docker-compose.yml logs app`. `make docker` runs `deploy/docker-compose.local.yml`, so that log command targets the wrong compose file for the SQLite path. Maybe add a separate magic-link extraction command for the local compose file, or tell users to copy it from the foreground `make docker` output.

My bad for missing those- addressed all three in the latest commits. Also caught a couple more 
1. while doing a fuller pass: base_url in INTEGRATE_CLAUDE_COWORK.md (should be server.public_url), 
2. stale claim in GOOGLE_OAUTH_SETUP.md that make setup prompts for Google OAuth credentials (it doesn't anymore), 
3. SQLITE_PATH default I had wrong. Should be clean now.

ericlevine:
Thank you! A few more minor points here:

docs/SETUP_LOCAL.md:213-217 still says Google services “require Google OAuth credentials from make setup.” That now contradicts the updated Step 3 text and the current cmd_setup.go flow, which only configures LLM/relay/telemetry. This should point users to docs/GOOGLE_OAUTH_SETUP.md or the Services UI instead.

docs/GOOGLE_OAUTH_SETUP.md:81-88 still tells users they can add Google credentials under services.google in config.yaml, but the current runtime does not load those fields into pkg/config.Config, and VaultOAuthProvider only reads GOOGLE_CLIENT_ID / GOOGLE_CLIENT_SECRET env vars or credentials stored in the system vault. Following this doc path would look configured but not work.

docs/GOOGLE_OAUTH_SETUP.md:92 says to start with make run. Given the PR now correctly documents that make setup writes ~/.clawvisor/config.yaml, this should probably use CONFIG_FILE="$HOME/.clawvisor/config.yaml" make run or refer back to SETUP_LOCAL.md so users don’t accidentally start against the repo-local default config.

I'm happy to do these as follow-ups if that's easier, too. Let me know!

adityajha2005:
> Thank you! A few more minor points here:
> 
> docs/SETUP_LOCAL.md:213-217 still says Google services “require Google OAuth credentials from make setup.” That now contradicts the updated Step 3 text and the current cmd_setup.go flow, which only configures LLM/relay/telemetry. This should point users to docs/GOOGLE_OAUTH_SETUP.md or the Services UI instead.
> 
> docs/GOOGLE_OAUTH_SETUP.md:81-88 still tells users they can add Google credentials under services.google in config.yaml, but the current runtime does not load those fields into pkg/config.Config, and VaultOAuthProvider only reads GOOGLE_CLIENT_ID / GOOGLE_CLIENT_SECRET env vars or credentials stored in the system vault. Following this doc path would look configured but not work.
> 
> docs/GOOGLE_OAUTH_SETUP.md:92 says to start with make run. Given the PR now correctly documents that make setup writes ~/.clawvisor/config.yaml, this should probably use CONFIG_FILE="$HOME/.clawvisor/config.yaml" make run or refer back to SETUP_LOCAL.md so users don’t accidentally start against the repo-local default config.
> 
> I'm happy to do these as follow-ups if that's easier, too. Let me know!

no worries, i have made the changes!

ericlevine:
Merged! Thank you for the PR!

PR #404, opened 2026-05-18
feat(adapters): add ClickUp via MCP

Myestery:
## Summary

Adds **ClickUp** as a connected service backed by ClickUp's hosted MCP
server (https://mcp.clickup.com/mcp, OAuth-only, public beta).

Replaces the original REST-based ClickUp adapter that earlier revisions
of this branch carried. ClickUp does not support PAT/API-token auth on
their MCP path, so a hand-rolled REST integration becomes redundant the
moment the MCP server can drive the same surface area natively.

The adapter is a 19-line spec:

- `transport: http`, `endpoint: https://mcp.clickup.com/mcp`
- `oauth: {}` — triggers the existing RFC 9728 / 8414 / 7591 discovery
  chain in `pkg/adapters/mcpadapter/discovery.go`; client_id/secret are
  registered dynamically and stored under `mcp.client.clickup`.
- `whoami: list_workspaces → teams[0].name` — mirrors the supabase.mcp
  pattern; the alias is the workspace name.

Tool list, parameter shapes, risk classification, and response handling
all come from the generic MCP middleware — no per-service Go code.

## Why MCP over REST

- ClickUp now ships and maintains the MCP server; we don't need to chase
  v2/v3 endpoint drift in adapter YAML.
- Auth is OAuth, so users avoid pasting a `pk_*` token into the vault.
- 381 lines of YAML + a service-specific httptest suite collapse to
  ~20 lines of config.

## Changes

- Add `internal/adapters/definitions/mcp/clickup.mcp.yaml`
- Add chain-context regex for ClickUp task IDs in
  `internal/intent/extractor.go` (works against MCP tool-result JSON
  exactly as it did against REST responses)
- Add `web/public/logos/clickup.png` (32x32 favicon)

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./pkg/adapters/{yamlloader,mcpadapter,yamlruntime}/...`
      passes
- [x] `go test ./internal/intent/... ./internal/adapters/...` passes
- [x] `go build ./cmd/clawvisor` succeeds
- [ ] Live OAuth: connect ClickUp, confirm workspace-name alias in
      `clawvisor services list`
- [ ] Live tool call: list workspaces / create + delete a task,
      confirm `task_id` chain-context is captured

## Follow-ups (out of scope)

- Intent eval cases for ClickUp once tool list is observed live
- Per-tool risk overrides if MCP defaults need tuning for write-class
  actions


ericlevine:
@Myestery Thank you so much for the submission! One thought -- it looks like ClickUp has a documented MCP server at https://mcp.clickup.com/mcp based on what I'm seeing [here](https://developer.clickup.com/docs/connect-an-ai-assistant-to-clickups-mcp-server). One thought would be whether it would be easier to use that instead of building the adapter directly. You can see an example of how relatively simple the MCP versions of adapters is [here](https://github.com/clawvisor/clawvisor/blob/main/internal/adapters/definitions/mcp/supabase.mcp.yaml), and it doesn't require as much long-term maintenance since it discovers the capabilities based on what the MCP server advertises. Would you mind giving that a try and see if it meets your needs?

Myestery:
> @Myestery Thank you so much for the submission! One thought -- it looks like ClickUp has a documented MCP server at https://mcp.clickup.com/mcp based on what I'm seeing [here](https://developer.clickup.com/docs/connect-an-ai-assistant-to-clickups-mcp-server). One thought would be whether it would be easier to use that instead of building the adapter directly. You can see an example of how relatively simple the MCP versions of adapters is [here](https://github.com/clawvisor/clawvisor/blob/main/internal/adapters/definitions/mcp/supabase.mcp.yaml), and it doesn't require as much long-term maintenance since it discovers the capabilities based on what the MCP server advertises. Would you mind giving that a try and see if it meets your needs?

Wow, will definitely check it out and update since they have an official MCP.
Though it's in beta and doesn't support deleting things, that should get updated soon and overall easier to manage from here

PR #412, opened 2026-05-19
fix(yamlruntime): propagate io.ReadAll and credentialFields errors

adityajha2005:
## Summary

- `pkg/adapters/yamlruntime/rest.go`: check error from `io.ReadAll` and return it instead of silently discarding
- `pkg/adapters/yamlruntime/graphql.go`: same fix in the GraphQL execution path
- `pkg/adapters/yamlruntime/adapter.go`: propagate `credentialFields` errors at both call sites (`Execute` and `FetchIdentity`) instead of ignoring them with `_`

## Test plan

- [x] `go build ./pkg/adapters/yamlruntime/...` passes (verified locally)
- [x] Run existing adapter tests: `go test ./pkg/adapters/...`
image

### DIFF
--- a/pkg/adapters/yamlruntime/adapter.go
+++ b/pkg/adapters/yamlruntime/adapter.go
@@ -124,8 +124,9 @@ func (a *YAMLAdapter) Execute(ctx context.Context, req adapters.Request) (*adapt
 	}
 
 	// Get credential fields for path and base-URL interpolation.
+	// Ignore parse errors when no credentials are provided (noauth adapters).
 	credFields, err := credentialFields(a.def.Auth, req.Credential)
-	if err != nil {
+	if err != nil && len(req.Credential) > 0 {
 		return nil, fmt.Errorf("%s: parsing credentials: %w", a.def.Service.ID, err)
 	}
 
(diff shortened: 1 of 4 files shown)


adityajha2005:
all checks passed finally

ericlevine:
Thanks for the fix, @adityajha2005! The io.ReadAll handling looks good, but I think the credential parsing change introduces a regression for OAuth adapters.

ValidateCredential currently accepts OAuth credentials that contain only a refresh_token, and buildOAuthClient supports refreshing those into an access token. With this PR, Execute/FetchIdentity call credentialFields(...) and return its error before the OAuth client can refresh. Since credentialFields goes through extractToken, refresh-token-only credentials now fail with credential missing token.

Could you update the credential-field parsing to be auth-aware? For auth.type: oauth2, it should not require an access token just to resolve base URL/path interpolation, or it should parse OAuth credentials separately and allow refresh_token-only credentials. A regression test for refresh-token-only OAuth execution would be helpful too.

adityajha2005:
> Thanks for the fix, @adityajha2005! The io.ReadAll handling looks good, but I think the credential parsing change introduces a regression for OAuth adapters.
> 
> ValidateCredential currently accepts OAuth credentials that contain only a refresh_token, and buildOAuthClient supports refreshing those into an access token. With this PR, Execute/FetchIdentity call credentialFields(...) and return its error before the OAuth client can refresh. Since credentialFields goes through extractToken, refresh-token-only credentials now fail with credential missing token.
> 
> Could you update the credential-field parsing to be auth-aware? For auth.type: oauth2, it should not require an access token just to resolve base URL/path interpolation, or it should parse OAuth credentials separately and allow refresh_token-only credentials. A regression test for refresh-token-only OAuth execution would be helpful too.

sure lemme fix this!

ericlevine:
Thank you! Merging.

PR #432, opened 2026-05-22
fix(adaptergen): isolate generated adapters per user

### LINKED ISSUE #301: [Critical] Cross-tenant adapter overwrite in generated adapter installation

devcool20:
## Summary

Closes #301

This fixes user-generated adapter isolation end to end.

Previously, installing a generated adapter could hot-load it into the shared adapter registry. In a multi-user/cloud deployment, that means a custom adapter generated by one user could become process-wide state and be visible/resolvable outside that user’s scope.

This PR makes generated adapters user-scoped through install, update, remove, catalog listing, activation, vault lookup, and execution paths.

## What Changed

- User-scoped adapter installs now hot-load into the per-user registry cache instead of the shared registry.
- User-scoped installs reject service IDs that collide with built-in/shared adapters.
- `Update` and `Remove` now resolve generated adapters through `GetForUser`, so cloud/user-scoped adapters can be managed after install.
- Added a user adapter listing hook to the registry so catalog/listing paths can discover user-only generated adapters when the service ID is not already known.
- Wired the cloud resolver to list DB-backed generated adapters for the current user and cache them per user.
- Updated service, welcome, skill catalog, task, approval, and gateway paths to use user-aware adapter/vault-key lookup where generated adapters may be involved.

## Why This Fully Addresses The Issue

The incomplete fix only handled part of the install path. This PR covers the missing surfaces called out in review:

- A generated adapter is not registered globally.
- Another user cannot resolve that adapter through `GetForUser`.
- Built-in adapter IDs cannot be overwritten by generated adapters.
- User-generated adapters still appear in `/api/services` and catalog-style listing paths after restart via `AllForUser`.
- Activation and execution-related paths use user-aware registry/vault-key resolution, so a listed user-generated adapter can actually be used by its owner.

## Regression Coverage

Added tests for:

- User-scoped generated adapter install does not touch the shared registry.
- The owning user can resolve the generated adapter.
- A different user cannot resolve it.
- Built-in/shared adapter ID collisions are rejected before save/hot-load.
- User-scoped generated adapter update works.
- User-scoped generated adapter removal evicts only the user-scoped adapter.
- `AllForUser` hydrates user-only adapters for catalog/listing paths.
- `AllForUser` does not allow a listed user adapter to override a global adapter.
- `/api/services` includes a user-generated adapter for the owning user.

## Validation

Targeted validation passed:

```text
go test ./pkg/adapters
go test ./internal/adaptergen
go test ./internal/api/handlers
go test ./pkg/clawvisor
go vet ./pkg/adapters ./internal/adaptergen ./internal/api/handlers ./pkg/clawvisor
git diff --check
```

Focused regression runs passed:

```text
go test ./pkg/adapters -run "Test(GetForUser|AllForUser)" -count=1 -v
go test ./internal/adaptergen -run "Test(Install_UserScoped|Update_UserScoped|Remove_UserScoped|Generate_EndToEnd)" -count=1 -v
go test ./internal/api/handlers -run "TestServicesListIncludesUserGeneratedAdapter" -count=1 -v
```

Relevant passing package output:

```text
ok      github.com/clawvisor/clawvisor/internal/adaptergen      (cached)
ok      github.com/clawvisor/clawvisor/internal/api/handlers    (cached)
ok      github.com/clawvisor/clawvisor/pkg/adapters             (cached)
ok      github.com/clawvisor/clawvisor/pkg/clawvisor            (cached)
```

Full suite was attempted on Windows:

```text
go test ./...
```

It is blocked by existing Windows/platform-specific failures unrelated to this PR, including Unix-only process APIs:

```text
internal/runtime/isolation/prune.go:153:17: undefined: syscall.Kill
internal/local/executor/exec.go:70:41: unknown field Setpgid in struct literal of type syscall.SysProcAttr
internal/local/executor/exec.go:229:23: undefined: syscall.Getpgid
internal/local/executor/exec.go:233:14: undefined: syscall.Kill
internal/local/executor/server.go:236:41: unknown field Setpgid in struct literal of type syscall.SysProcAttr
```

There are also unrelated Windows file permission expectation failures:

```text
vault key permissions = 666, want 0600
raw log mode=0666, want 0600
adjudication debug dir ... must be owner-only (0700)
```

The packages touched by this PR pass their targeted tests and vet checks locally.

### DIFF
--- a/internal/adaptergen/generator.go
+++ b/internal/adaptergen/generator.go
@@ -27,32 +27,32 @@ type Generator struct {
 	genClient  *llm.Client // high max_tokens for YAML generation
 	riskClient *llm.Client // lower max_tokens for risk classification JSON
 	registry   *adapters.Registry
-	store       AdapterStore
-	userID      string // scopes cache operations; empty for local single-user mode
-	logger      *slog.Logger
+	store      AdapterStore
+	userID     string // scopes cache operations; empty for local single-user mode
+	logger     *slog.Logger
 }
 
 // GenerateResult contains the output of a generation attempt.
 type GenerateResult struct {
-	ServiceID   string         `json:"service_id"`
-	DisplayName string         `json:"display_name"`
-	Description string         `json:"description,omitempty"`
-	BaseURL     string         `json:"base_url"`
-	AuthType    string         `json:"auth_type"`
-	YAML        string         `json:"yaml"`
+	ServiceID   string          `json:"service_id"`
+	DisplayName string          `json:"display_name"`
+	Description string          `json:"description,omitempty"`
+	BaseURL     string          `json:"base_url"`
+	AuthType    string          `json:"auth_type"`
+	YAML        string          `json:"yaml"`
 	Actions     []ActionPreview `json:"actions"`
-	Warnings    []string       `json:"warnings,omitempty"`
-	Installed   bool           `json:"installed"`
+	Warnings    []string        `json:"warnings,omitempty"`
+	Installed   bool            `json:"installed"`
 }
 
 // ActionPreview is a user-friendly summary of a generated action.
 type ActionPreview struct {
-	Name        string        `json:"name"`
-	DisplayName string        `json:"display_name"`
-	Method      string        `json:"method,omitempty"`
-	Path        string        `json:"path,omitempty"`
-	Category    string        `json:"category"`
-	Sensitivity string        `json:"sensitivity"`
+	Name        string         `json:"name"`
+	DisplayName string         `json:"display_name"`
+	Method      string         `json:"method,omitempty"`
+	Path        string         `json:"path,omitempty"`
+	Category    string         `json:"category"`
+	Sensitivity string         `json:"sensitivity"`
 	Params      []ParamPreview `json:"params,omitempty"`
 }
 
@@ -158,7 +158,7 @@ func (g *Generator) Install(ctx context.Context, yamlContent string) (*GenerateR
 // Update regenerates an existing adapter from new source material.
 // The old adapter is replaced in-place.
 func (g *Generator) Update(ctx context.Context, serviceID string, src Source) (*GenerateResult, error) {
-	if _, ok := g.registry.Get(serviceID); !ok {
+	if _, ok := g.registry.GetForUser(ctx, serviceID, g.userID); !ok {
 		return nil, fmt.Errorf("adapter %q not found in registry", serviceID)
 	}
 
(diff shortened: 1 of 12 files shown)


ericlevine:
This looks great. Thanks for the PR.

devcool20:
> This looks great. Thanks for the PR.

happy to help, always!

PR #433, opened 2026-05-22
refactor(onboarding): upgrade GetStarted UI and fix scroll tracking

Satyamyaduvanshi:
## What does this PR do?
This PR upgrades the UI and layout of the `GetStarted` page to make the onboarding flow cleaner and more robust, alongside fixing a scrolling bug that prevented the sidebar from tracking correctly.

### 1. Get Started UI Upgrades
* **Grid Layout:** Transitioned the "Connect a service/agent" lists into a 2-column grid (`md:grid-cols-2`) to better utilize horizontal screen real estate.
* **Asset Optimization:** Replaced large inline SVGs for third-party services with native `` tags pointing to `/logos/` to keep the React component clean. Added `dark:invert` so dark-colored logos (like GitHub) remain visible in dark mode.
* **Component Styling:** Upgraded the "Three layers of control" section with a branded background and flexbox shield icon to visually emphasize it as a core security concept.

### 2. Scroll Tracking Fixes
* **Custom Scroll Spy:** Replaced the standard `IntersectionObserver` with a custom `useScrollSpy` window scroll listener. This prevents the "short page" bug where bottom sections couldn't trigger the active state if the screen was taller than the content.
* **Restored `position: sticky`:** Removed `overflow-auto` from the `<main>` tag in `Dashboard.tsx`. Because the parent uses `min-h-screen`, this vertical overflow wasn't necessary for page scrolling, but it created a hidden scroll container that broke `position: sticky` for the new right-hand sidebar.

## Testing
- [x] Ran `cd web && npx tsc --noEmit && npm run build` locally successfully.
- [x] Verified grid layout responsiveness.
- [x] Verified sidebar scroll tracking across different viewport heights.

## Type of change
- [x] `refactor`: UI/UX enhancements and layout polish
- [x] `fix`: Resolves sticky sidebar and scroll tracking

## Before 
Screenshot_20260522_181651

## After 
Screenshot_20260522_182846


Satyamyaduvanshi:
Thanks for the thorough review, @ericlevine (and Forge)! Glad you like the new UI direction.

I've pushed a new commit that addresses all the feedback and marked this as ready for review:

Added the missing EOF newline to GetStarted.tsx.

Cleaned up the trailing whitespace in index.css.

Swapped sectionIds.join(',') to JSON.stringify(sectionIds) in the useEffect dependency array.

Regarding the layout and logos:

Logos: Confirmed! The SVGs already exist in the /public/logos/ directory so they render perfectly without silently failing.

Overflow: I did a quick audit of the other dashboard views (Tasks, Activity, etc.). Because we preserved min-w-0 on the <main> tag, the wide tables and code blocks still correctly trigger their own horizontal scrollbars without breaking the global layout.

Satyamyaduvanshi:
@ericlevine waiting for reply!

ericlevine:
Thanks for the thoughtful upgrade — the new card layout, the `/logos/*` swap with `dark:invert`, and the "Three layers of control" callout all look great. A few things I'd like to sort out before merging:

**Higher-impact items:**

1. **`Dashboard.tsx` `overflow-auto` → `overflow-x-clip` is a layout-wide change.** That `<main>` is the shell for every dashboard page (Audit, Services, AdapterGen, Inspector, etc.), not just GetStarted. Moving the scroll from the inner element to the window changes scroll-restoration behavior across route navigations, and `overflow-x-clip` *hides* horizontal overflow rather than scrolling it. Could you verify the other long pages still behave correctly, and update the test plan to cover them?

2. **Heading hierarchy in `SetupStepCard`.** The card title changed from `<h3>` to `<h2>`, which means two `<h2>`s now nest under the section's `<h2>` ("Complete your workspace setup"). Other sections (`SuggestionsSection`, `YourSetupSection`) keep the h2-section / h3-card pattern. Could we revert this one to `<h3>` for consistency and a11y?

3. **Empty `<p>` in the right-rail header.** The element contains only the hamburger SVG with no visible text — looks like the label (e.g. "On this page") got dropped. Could you either add the label back or swap to a non-text element?

**`useScrollSpy` polish:**

- `JSON.stringify(sectionIds)` as a dep is a smell; `useMemo` on the array upstream (or a ref-based equality check) would be cleaner.
- The `setTimeout(..., 100)` to wait for paint is fragile under slower renders. `requestAnimationFrame` (or just attaching the listener immediately and calling `handleScroll()` once) is more reliable.
- Worth adding a `resize` listener — viewport changes shift which element crosses the detection line.

**Smaller things:**

- The file-header "RECENT ARCHITECTURAL CHANGES & DEPRECATIONS" block tends to rot fast and isn't the pattern elsewhere in the codebase. The PR description already captures this — could we drop the in-file version?
- The hand-drawn `ClaudeCodeIcon` / `ClaudeDesktopIcon` / `OpenClawIcon` SVGs don't match the actual brand marks, and the two Claude ones are visually near-identical. Could we either use real logos via the `/logos/` pattern you already adopted for services, or skip icons here (the previous `PopularAgent` had none)?
- `scroll-behavior: smooth` on `html` is global; if it's only needed for the GetStarted anchor links, scoping it to that container would avoid surprising behavior elsewhere.
- Tiny one: the "You ask" step now interpolates the prompt without surrounding quotes — was that intentional? It used to read more obviously as a quotation.
- The right rail is `xl:block` (≥1280px), so a chunk of laptop users won't see it. Worth considering `lg:` if the in-page nav is part of the value here.

**One heads-up on direction:** we're about to make the LLM proxy the default onboarding path for open agent harnesses (Claude Code, OpenAI-SDK-based agents, etc.), which will shift a lot of the "connect an agent" prominence away from this page. The HTTP gateway flow that GetStarted currently centers on stays first-class for closed harnesses and backend pipelines, but it'll be one of two paths rather than *the* path. None of this blocks the PR — the improvements here are still wins — but worth knowing before you sink more time into deeper restructuring of this surface. Happy to share more context if useful.

Happy to chat through any of these if I'm misreading the intent — and again, the visual direction is a real step up.

Satyamyaduvanshi:
@ericlevine

ericlevine:
Looks great! Thank you. Will merge now. Appreciate your patience on this.

PR #446, opened 2026-05-25
Fixed Drive query escape (sheets/actions.go) issue

Unmesh100:
<!-- lightsprint-task-links -->
## 🔗 Lightsprint Task
[CLA-1](https://app.lightsprint.ai/repos/EEkzD_teBDB9tncmjYfvk/tasks/1)

**Assignee:** Eric Levine

**Deployment:** production
<!-- /lightsprint-task-links -->

Bug Fixes: Security Vulnerabilities                                                                                                              
                                                                                                                                                 
  1. Telegram Callback Authentication Bypass                                                                                                       
                                                            
  File: internal/notify/telegram/polling.go:305                                                                                                    
                                                                                                                                                   
  The bug: The auth check compared cq.From.ID (int64) to ps.chatID (string) by converting the integer to a string via fmt.Sprintf. This meant
  comparing e.g. "123456789" to "123456789" — but since ps.chatID is a string from user config (not the numeric user ID), this comparison always
  failed, rejecting every callback query as unauthorized.

  The fix: Parse ps.userID as int64 and compare integers directly.

  // Before (broken — always rejects)
  if cq.From.ID != 0 && fmt.Sprintf("%d", cq.From.ID) != ps.chatID {
      n.answerCallbackQuery(ctx, ps.botToken, cq.ID, "Not authorized.")
      return
  }

  // After (correct — compares numeric user IDs)
  if cq.From.ID != 0 {
      psUserID, err := strconv.ParseInt(ps.userID, 10, 64)
      if err != nil || cq.From.ID != psUserID {
          n.answerCallbackQuery(ctx, ps.botToken, cq.ID, "Not authorized.")
          return
      }
  }

  ---
  2. Google Drive Query Injection

  File: internal/adapters/google/sheets/actions.go:54

  The bug: User-provided query was embedded directly into the Drive API query with single quotes: name contains 'userinput'. The character
  allowlist blocked Drive operator keywords but allowed parentheses — so a payload like ') or trashed=false or name contains (' could break out of
  the string literal and inject arbitrary query operators.

  The fix: Escape embedded single quotes with '' (Drive API's escape convention) before interpolation.

  // Before (vulnerable)
  driveQ += fmt.Sprintf(" and name contains '%s'", query)

  // After (escaped)
  driveQ += fmt.Sprintf(" and name contains '%s'", strings.ReplaceAll(query, "'", "''"))

  ---
  3. Go Template Injection

  File: pkg/adapters/yamlruntime/template.go:34

  The bug: template.Parse() was given a FuncMap that included the index function (a Go template built-in). index is the primary injection vector —
  it allows accessing arbitrary map keys, including ones that trigger template.HTML construction internally, which bypasses auto-escaping and can
  execute arbitrary template content.

  The fix: Replace the full FuncMap with an explicit allowlist of safe functions (len, upper, print). Block index by not exposing it.

  // Before (exposes index — injection vector)
  t, err := template.New("summary").Funcs(summaryFuncs).Parse(tmplStr)

 // After (restricted — no index, no custom funcs)
  t, err := template.New("summary").
      Funcs(template.FuncMap{
          "len":   summaryFuncs["len"],
          "upper": summaryFuncs["upper"],
      }).
      Funcs(template.FuncMap{
          "print": func(a ...any) string { return fmt.Sprint(a...) },
      }).
      Parse(tmplStr)


Unmesh100:
@ericlevine can you please review it, thanks.

ericlevine:
Thanks for putting this together. I went through each of the three changes carefully and want to share findings before we merge. Two of them look like net regressions to me, and the third is good but worth reframing.

**1. Telegram callback auth (`polling.go`) — please revert**

`ps.userID` is the Clawvisor-internal user ID (an opaque string passed through from `req.UserID` at the call sites in `telegram.go`), not a numeric Telegram ID. `strconv.ParseInt` will return an error on every real session, the `err != nil` branch will fire, and every callback button tap will be rejected as "Not authorized." This breaks the inline keyboard flow.

The original comparison was against `ps.chatID`, which is the numeric Telegram chat ID stored as a string. For 1:1 chats with the bot (the common case) `chat_id == from.id` on the Telegram API, so `fmt.Sprintf("%d", cq.From.ID) != ps.chatID` was correct. If we want to tighten this further, the right move is to plumb the Telegram user ID through as its own field on `pollingSession` — not to repurpose the Clawvisor `userID`.

**2. Template "fix" (`yamlruntime/template.go`) — please revert**

A few things don't line up here:

- This file imports `text/template`, not `html/template`. There's no HTML auto-escaping in play, and `index` doesn't construct `template.HTML` values.
- `index` is a built-in template function — omitting it from `FuncMap` doesn't disable it. The in-code comment ("The built-in index function is allowed") actually contradicts the PR description ("Block index by not exposing it").
- The change adds a new `print` function on top of the original `len`/`upper`, so it widens the function surface rather than restricting it.
- `.Delims("{{", "}}")` is a no-op (those are the defaults), and splitting `Funcs` across two calls just makes the precedence harder to reason about.

If there's a real concern here, I'd like to first nail down where `tmplStr` originates. If it comes from operator-authored YAML adapter definitions in this repo, this isn't an injection surface at all. If it can come from end-user input, we'd need a meaningfully different design (depth limits, no reflection-based calls, untrusted data never fed as the template *string*).

**3. Drive query escape (`sheets/actions.go`) — keep, but reframe**

This one I'd like to keep. The escape itself is correct — `''` is Drive's documented convention. One nuance worth surfacing in the PR description though: the existing `driveQueryAllowlist` regex doesn't include `'`, so the example payload (`') or trashed=false or name contains ('`) is actually rejected at line 51 before reaching the format string. That makes this defense-in-depth against a future allowlist widening, not a fix for a live exploit. Honest framing + still worth merging.

It'd be great to add two small tests in `actions_test.go`:

1. A test asserting that a query containing `'` fails the allowlist check (locks in the actual defense).
2. Extract the escape into a tiny helper (`escapeDriveLiteral`) and unit-test it on `it's`, `') or x=('`, etc.

**Suggested path forward**

Could we split this into separate PRs? Land just the Drive escape (with the tests and reframed description) and revert the other two. Happy to discuss a fix for the Telegram auth check if there's a specific scenario where the chatID-based comparison is failing — would want to see a repro before changing the semantics.

Unmesh100:
is it fine now @ericlevine ? otherwise i would love to make separate PRs after closing this.

ericlevine:
Hi @Unmesh100, it's looking better, thank you! Can you update the PR title / description to match the work that's now represented here?

Unmesh100:
Done @ericlevine 

Let me know if you want me to contribute on any new features or issues.

ericlevine:
Thanks for the PR! I'll let you know if anything comes to mind.

PR #465, opened 2026-05-28
fix(llmproxy): preserve lite-proxy audit child rows

### LINKED ISSUE #464: Lite-proxy audit child events can be dropped by request-level
# Lite-proxy Audit Child Events Can Be Dropped by Request-Level Deduplication

## Problem

Lite-proxy can emit multiple audit events during a single upstream LLM request. This can happen when one model response contains multiple tool uses, or when the proxy records child/history events such as:

* approval releases
* inline task approvals
* auto-approvals
* skipped sibling tools
* resolver swaps

Today these rows can share the same canonical audit scope:

```sql
(user_id, request_id, COALESCE(task_id, ''))
```

Because canonical audit rows are deduplicated on that scope, only one row may survive for a given request/task. Later lite-proxy child events can conflict and be dropped.

---

## Impact

This can make audit history and dashboard task activity incomplete.

Examples:

* A model emits multiple tool calls in one response, but only one `runtime.tool_use` audit row is persisted.
* Coalesced approval handling records only part of the sibling tool-use story.
* Resolver swap retries or repeated approval-related audit events may be missing.
* The dashboard activity feed can show an incomplete or misleading timeline for a task.

This is especially risky for the LLM proxy launch/demo because task activity depends on accurate lite-proxy audit rows.

---

## Root Cause

The audit dedup key is currently request-level, but lite-proxy has both:

1. request-level audit rows
2. child event rows

Request-level rows should remain deduplicated by request ID for idempotency.

Child event rows, however, need to preserve the original `request_id` while also having their own event-level uniqueness.

Changing the stored `request_id` for child rows would avoid the index collision, but that can break consumers that rely on the original upstream request identifier for:

* lookup
* grouping
* debugging

---

## Proposed Fix

Add an optional `dedup_key` column to audit rows.

Use the canonical uniqueness rule:

```sql
(user_id, COALESCE(dedup_key, request_id), COALESCE(task_id, ''))
WHERE deduped_of IS NULL
```

### Behavior

* Existing/request-level rows keep `dedup_key = NULL`
* Existing dedup semantics remain unchanged for request-level audit rows
* Lite-proxy child/history rows keep the original `request_id`
* Lite-proxy child/history rows set a per-event `dedup_key`
* Request idempotency lookup should continue to use request-level rows only
* Request-based audit lookup should still be able to return child rows

---

## Lite-proxy Child Rows That Should Use Event-Level Dedup

The following child/history events should use event-level deduplication:

* tool-use inspection
* approval release
* inline task approval
* inline task auto-approval
* continuation skipped sibling tools
* resolver swap

---

### Persistence

* Multiple `runtime.tool_use` rows from the same `request_id` and `task_id` can persist
* Repeated lite-proxy child/history events from the same request can persist
* Request-level duplicate rows are still deduplicated
* Child rows preserve the original `request_id`

### Migration / Compatibility

* Existing audit rows do not require a data backfill
* SQLite and Postgres migrations preserve old uniqueness semantics for rows with `dedup_key IS NULL`

### Lookup Behavior

* Request-based audit lookups do not skip child rows

### Testing

Focused regression tests cover:

* multi-tool rows
* repeated child events
* request-level dedup preservation
* lookup behavior

---

## Summary

The current request-level deduplication strategy is too coarse for lite-proxy child/history events.

Introducing an optional `dedup_key` allows:

* request-level rows to remain idempotent
* child/history rows to persist independently
* original `request_id` values to remain intact for debugging and lookup compatibility

This preserves backward compatibility while enabling complete and accurate task activity history for lite-proxy workflows.

devcool20:
## Summary

Fixes #464

Lite-proxy can emit several audit events under the same upstream LLM request, including tool-use inspection rows, approval release rows, inline task approval rows, auto-approval rows, skipped sibling rows, and resolver swap rows.

Those rows were previously canonical audit rows sharing the same `(user_id, request_id, task_id)` scope. Because the audit table dedup index only allowed one canonical row for that scope, later child events from the same request could be silently dropped.

This adds an optional `dedup_key` column for canonical audit rows. Request-level rows continue to leave `dedup_key` empty and preserve the existing request-id dedup behavior. Lite-proxy child/history rows keep the original `request_id` for lookup/grouping compatibility, but receive a per-event `dedup_key` so multiple legitimate child audit events can coexist under one request.

## What changed

- Added nullable `dedup_key` support to audit entries.
- Updated SQLite and Postgres audit insert/scan paths.
- Added SQLite and Postgres migrations for the new dedup key index.
- Preserved existing request-id uniqueness semantics for rows where `dedup_key` is null.
- Updated lite-proxy child audit emitters to set per-event dedup keys:
  - tool-use inspection
  - approval release
  - inline task approval
  - inline task auto-approval
  - continuation skipped sibling tools
  - resolver swap
- Kept `request_id` unchanged on child rows so existing request-based consumers do not break.
- Kept request-level idempotency lookup scoped to `dedup_key IS NULL`.
- Ensured request-based audit lookup can still return child rows.
- Relaxed the synthetic approval response regression test to compare semantic JSON equality instead of byte-identical JSON output.

## Test plan

- `go test ./internal/runtime/llmproxy -run "TestAuditEmitter_LogToolUseInspected|TestAuditLog_RuntimeToolUsePreservesRepeatedChildEvents|TestAuditLog_LiteProxyChildrenPreserveEventHistory|TestAuditLog_RequestLookupIncludesChildDedupRows|TestSyntheticApprovalToolUses_SingletonMatchesLegacy|TestPostprocess_CoalesceAuditDoesNotEmitMisleadingAllow|TestPostprocess_CoalescePendingAuditSurfacesApprovalTrigger" -count=1`
- `go test ./pkg/store ./pkg/store/sqlite -count=1`
- `go test ./internal/api/handlers -count=1`
- `git diff --check`


devcool20:
hey @ericlevine, have a look at this.
issue #464  was a blocker and big critical issue (dashboard's activity feed relies on these audit logs)

ericlevine:
Thanks for the PR, @devcool20! A few smaller follow-ups from re-reading the dedup change end-to-end:

**Stale comments around the precedence flip**

The PR flipped `FindDedupCandidate` from pre-task-first to exact-task-first, but four docstrings still describe the old world where `GetAuditEntryByRequestIDAndTask` "inverts" FindDedupCandidate's precedence. Now they share the same task-tier precedence (the only remaining difference is within-tier ordering: the getter takes newest, the dedup takes oldest).

- `pkg/store/store.go:140-144` — interface doc for `GetAuditEntryByRequestIDAndTask`
- `pkg/store/sqlite/store.go:1018-1022` — implementation doc
- `pkg/store/postgres/store.go:853-855` — implementation doc
- `internal/api/handlers/gateway.go:1328-1331` — `lookupAuditByRequestID` caller-side comment

Worth fixing in the same change so the model stays consistent across the interface, the two backends, and the consumer.

**ErrConflict warn-spam on replay-safe child events**

`LogToolUseInspected`, `LogApprovalRelease`, `LogInlineTaskApproved`, `LogInlineTaskAutoApproved`, `LogContinuationSkippedSiblingTools`, and `LogResolverSwap` all derive a deterministic `dedup_key` from `(request_id, tool_use_id, …)`. When the same event fires twice — a retry, a poll-driven replay, etc. — the new partial unique index correctly rejects it as `store.ErrConflict`. That's the dedup working as designed, but each emitter still logs `"… audit failed"` at warn level.

Consider gating the warn behind `!errors.Is(err, store.ErrConflict)` on those six sites so replays don't fill the operational log with successful no-ops. (Leave `LogEndpointCall` as-is — its row is request-level, no `dedup_key`, so an `ErrConflict` there really does mean a race past the gateway's `FindDedupCandidate` check, which is worth a warn per the `LogAudit` contract.)

ericlevine:
Looks great. Thanks for the contribution, @devcool20!

PR #431, opened 2026-05-21
feat: add Google Sheets adapter with OAuth integration 

rakshityadav1868:
fixes #430 
Adds a new `google.sheets` adapter with Google OAuth integration and MVP spreadsheet actions.

### Features

- Added `google.sheets` adapter
- Reused existing Google OAuth + vault credential flow
- Added service catalog entry via `google_sheets.yaml`
- Wired Go overrides in `defaults.go`
- Added validation, metadata, and semantic response formatting

### Supported Actions

- `list_spreadsheets`
- `get_spreadsheet`
- `read_range`
- `append_rows`
- `update_cells`
- `create_spreadsheet`

### Notes

- Shares the existing Google OAuth connection model using:
  - `vault_key: google`
  - `scope_merge: true`

### PREVIEW

https://drive.google.com/drive/u/0/folders/1oWeN3vWqZSxoaaEK4mAErPi3osRlm2aE

PR #469, opened 2026-05-28
feat: Add builtin fact extraction patterns for Google Sheets spreadsheet IDs

rakshityadav1868:
## Summary

refrence #431 
Adds builtin extraction patterns for google.sheets so spreadsheetId values are automatically extracted into chain facts.

This follows the same approach already used for sibling Google services like:
- Gmail (message_id, thread_id)
- Drive (file_id)
- Calendar (event_id)
- Contacts (phone_number, email_address)

## Changes

- Added a google.sheets case inside builtinPatternsRaw()
- Extracts:
  - spreadsheetId → file_id

## Notes

This is a small non-blocking improvement based on mentor feedback for service-specific fact extractors in internal/intent/extractor.go.

### DIFF
--- a/internal/intent/extractor.go
+++ b/internal/intent/extractor.go
@@ -855,7 +855,7 @@ func builtinPatternsRaw(service, action string) []extractionPattern {
 		)
 	case "google.sheets":
 		return append(generic,
-			extractionPattern{FactType: "file_id", Regex:    `"spreadsheetId":\s*"([^"]+)"`},
+			extractionPattern{FactType: "file_id", Regex: `"spreadsheet_id":\s*"([^"]+)"`},
 		)
 	case "google.calendar":
 		// Google Calendar event IDs are lowercase base32 (5-1024 chars),
--- a/internal/intent/extractor.go
+++ b/internal/intent/extractor.go
@@ -853,6 +853,10 @@ func builtinPatternsRaw(service, action string) []extractionPattern {
 		return append(generic,
 			extractionPattern{FactType: "file_id", Regex: `"id":\s*"([^"]+)"`},
 		)
+	case "google.sheets":
+		return append(generic,
+			extractionPattern{FactType: "file_id", Regex:    `"spreadsheetId":\s*"([^"]+)"`},
+		)
 	case "google.calendar":
 		// Google Calendar event IDs are lowercase base32 (5-1024 chars),
 		// optionally prefixed with "_" for imported events, with an optional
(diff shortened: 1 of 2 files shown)


rakshityadav1868:
hi @ericlevine could u take a look

ericlevine:
Thanks for picking this up — appreciate the follow up here to match the pattern alongside the existing Drive/Calendar/Contacts.

One blocker before this can land: the regex won't actually match anything in production. The `spreadsheetId` (camelCase) key only exists on the unexported request/response structs *inside* the sheets adapter that decode the upstream Google API. Before the payload reaches the extractor, it goes through `json.Marshal(*adapters.Result)` in `gateway.go:1841`, and the exported result types all tag the field as snake_case:

```go
// internal/adapters/google/sheets/actions.go
SpreadsheetID string `json:"spreadsheet_id"`   // L96, 149, 209, 275, 340
```

The Drive analogue works because Drive's emitted JSON tag is literally `"id"` — same key in and out. For Sheets the public shape is different from the upstream API shape, so the regex needs to target the public one:

```go
case "google.sheets":
    return append(generic,
        extractionPattern{FactType: "file_id", Regex: `"spreadsheet_id":\s*"([^"]+)"`},
    )
```

Two small follow-ups while you're in there:

1. **Add a test row.** `extractor_test.go` has a nice table at L295–297 with one entry per service. A `google.sheets` row with a realistic post-marshal payload (e.g. `{"spreadsheet_id":"1BxR7a3mNpQ9vK2wL5sYcTfDgHjKlMnO"}`) would have caught the casing mismatch and matches the convention for the sibling services.
2. **Whitespace nit.** `Regex:    "..."` has four spaces after the colon vs. one space everywhere else in the function — gofmt happens to tolerate it on a single-line literal, but it reads as a typo.

Using `file_id` (rather than introducing `spreadsheet_id` as a new fact type) is the right call — a spreadsheet *is* a Drive file, so a Drive lookup and a Sheets call can satisfy each other's IDs. And `list_spreadsheets` correctly falls through to the default branch since it returns Drive-shaped `{"id":...}` items.

Lastly, it would be great if you could make sure you smoke test this with the real Google Sheets adapter to make sure it's completely working end-to-end. Once the key is fixed, a test is added and you've confirmed that it works with a live adapter, this looks good to go. Thanks again!

rakshityadav1868:
@ericlevine Fixed all three things:

- Changed the regex key from `spreadsheetId` to `spreadsheet_id` — you were right, the post-marshal shape uses snake_case. Good catch on the casing mismatch between the upstream API and the exported types.
- Added a `google.sheets` test row using `{"spreadsheet_id":"1BxR7a3mNpQ9vK2wL5sYcTfDgHjKlMnO"}`. All tests pass including the per-service table. Would have caught this immediately.
- Cleaned up the whitespace (four spaces after `Regex:` down to one).

Tested the regex against a realistic payload and it matches correctly now. Running the full test suite against the sheets adapter next to confirm end-to-end. Thanks for the detailed review — the explanation of how the payload flows through `json.Marshal` was really helpful.

PR #470, opened 2026-05-28
fix(llmproxy): expire pending inline tasks on rollback

devcool20:
## Summary

Updates the inline-task rollback fix for the new PR #472 task-based architecture.

PR #472 replaced the temporary `pending_approvals` mirror approach with first-class `store.Task` rows at `status = pending_approval` and `approval_source = inline_chat`.

This PR now handles the remaining atomicity gap in that architecture: if proxy-lite creates a pending inline-chat task during response evaluation, but the turn later fails before its approval hold is safely committed, the task is expired so it does not remain as a zombie pending task in the dashboard.

## Changes

- Add rollback for buffered pending inline tasks on fail-closed postprocess exits.
- Expire pending inline tasks when:
  - `rewriter.Rewrite(...)` fails after evaluation.
  - `replayBufferedHolds(...)` fails before cache holds are committed.
- Use `ExpireInlineTask`, not `DenyInlineTask`, because these are operational rollbacks, not user denials.
- Add coverage proving a task created through the real `TasksHandler.CreatePendingInlineTask` path is moved to `expired` when replaying the cache hold fails.
- Add focused coverage for rollback helper filtering.

## Before / After

Before:

```text
proxy-lite creates store.Task:
  status = pending_approval
  approval_source = inline_chat

then cache hold replay fails

result:
  response fails
  cache hold is absent
  task remains pending_approval
  dashboard shows a task that chat can no longer resolve
After:

proxy-lite creates store.Task:
  status = pending_approval
  approval_source = inline_chat

then cache hold replay fails

rollback:
  pending inline task is expired via ExpireInlineTask

result:
  response fails closed
  cache has no orphaned hold
  dashboard has no zombie pending task
Test Plan
go test ./internal/api/handlers -run "TestPostprocess_ReplayFailureExpiresPendingInlineTask|Test.*InlineTask" -count=1
go test ./internal/runtime/llmproxy -run "TestReplayBufferedHolds|TestCleanupEvictedInlineTask|TestRollbackBufferedPendingTasks|TestPostprocess_CoalescedHoldEvictionExpiresInlineTask" -count=1
go test ./pkg/store ./pkg/store/sqlite -count=1
git diff --check HEAD~1..HEAD

### CODE REVIEW

**Review by devcool20 — COMMENTED** [review]
all the issues flagged were solved with the latest commit

**Comment by ericlevine** [comment]
Thanks for the iteration on this — the `failClosed` factoring is clean and the `Expire`-vs-`Deny` choice (with the explicit test guard) is the right semantics. A few things before this lands:

**1. Trace rollback expire failures.**
`rollbackBufferedPendingTasks` swallows errors from `ExpireInlineTask`:
```go
_ = pendingCreator.ExpireInlineTask(rollbackCtx, h.Pending.PendingTaskID, h.Pending.UserID)
```
The sibling helper `cleanupEvictedInlineTask` emits an `inline_task.evicted_expire_failed` trace event in the same situation (and `TestCleanupEvictedInlineTask_TracesExpireFailure` pins it). Without a corresponding event here, a rollback that itself fails leaves a zombie task with no breadcrumb. Suggest mirroring that pattern with something like `inline_task.rollback_expire_failed`.

**2. Add an end-to-end test for the rewriter-error branch.**
`TestPostprocess_ReplayFailureExpiresPendingInlineTask` covers the replay-failure `failClosed` site through the real `TasksHandler.CreatePendingInlineTask`, which is great. The rewriter-error site (line 929) only has the helper-level `TestRollbackBufferedPendingTasks_ExpiresOnlyInlineTaskHolds` behind it. A parallel `TestPostprocess_RewriterFailureExpiresPendingInlineTask` would keep the two fail-closed paths symmetric under CI and catch anyone deleting the `failClosed(...)` wrap on the rewriter site.

**3. Tiny: note the sequential-with-timeout assumption.**
The rollback is sequential with a 5s per-call timeout. Today the inline-task intercept produces at most one pending task per turn so this is a non-issue, but if that ever changes the failure path silently degrades to 5N seconds. A one-line comment ("currently ≤1 pending inline task per turn; parallelize if that changes") would protect the assumption.

Once (1) and (2) are addressed I think this is good to merge.

devcool20:
> Thanks for the iteration on this — the `failClosed` factoring is clean and the `Expire`-vs-`Deny` choice (with the explicit test guard) is the right semantics. A few things before this lands:
> 
> **1. Trace rollback expire failures.** `rollbackBufferedPendingTasks` swallows errors from `ExpireInlineTask`:
> 
> ```go
> _ = pendingCreator.ExpireInlineTask(rollbackCtx, h.Pending.PendingTaskID, h.Pending.UserID)
> ```
> 
> The sibling helper `cleanupEvictedInlineTask` emits an `inline_task.evicted_expire_failed` trace event in the same situation (and `TestCleanupEvictedInlineTask_TracesExpireFailure` pins it). Without a corresponding event here, a rollback that itself fails leaves a zombie task with no breadcrumb. Suggest mirroring that pattern with something like `inline_task.rollback_expire_failed`.
> 
> **2. Add an end-to-end test for the rewriter-error branch.** `TestPostprocess_ReplayFailureExpiresPendingInlineTask` covers the replay-failure `failClosed` site through the real `TasksHandler.CreatePendingInlineTask`, which is great. The rewriter-error site (line 929) only has the helper-level `TestRollbackBufferedPendingTasks_ExpiresOnlyInlineTaskHolds` behind it. A parallel `TestPostprocess_RewriterFailureExpiresPendingInlineTask` would keep the two fail-closed paths symmetric under CI and catch anyone deleting the `failClosed(...)` wrap on the rewriter site.
> 
> **3. Tiny: note the sequential-with-timeout assumption.** The rollback is sequential with a 5s per-call timeout. Today the inline-task intercept produces at most one pending task per turn so this is a non-issue, but if that ever changes the failure path silently degrades to 5N seconds. A one-line comment ("currently ≤1 pending inline task per turn; parallelize if that changes") would protect the assumption.
> 
> Once (1) and (2) are addressed I think this is good to merge.

addressed all the issues, i think we are good to go 👍

ericlevine:
Thank you!

PR #472, opened 2026-05-28
feat(runtime): surface inline-chat pending tasks via store.Task

ericlevine:
## Summary

Inline-chat task approvals (`POST /api/control/tasks?surface=inline`) now create a real `store.Task` at `pending_approval` instead of living only in the llmproxy in-process cache. The dashboard's existing Tasks page renders the pending row automatically; the cache hold remains the in-process resolution anchor.

**Why**: the previous behavior left the dashboard blind to a chat-bound approval-in-flight (the user had no durable "this is waiting on me" surface). A prior attempt mirrored the cache hold into `pending_approvals`, but that conflated two different lifecycles and shipped rows the dashboard couldn't actually resolve. This PR instead reuses the existing pending-task surface — same Task model, same Tasks page, same audit trail — with a marker (`ApprovalSource="inline_chat"`) so the dashboard's Approve/Deny endpoints refuse with `409 INLINE_CHAT_BOUND` and the chat reply is authoritative.

## Flow

1. Model emits `POST /api/control/tasks?surface=inline`.
2. Intercept validates, runs risk assessment, tries the auto-approve gate (unchanged).
3. On gate refusal: lands a `store.Task` at `pending_approval` with `approval_source="inline_chat"` plus a canonical `approval_records` row at `surface="inline_chat"`, `status="pending"`. The Tasks page renders it.
4. Cache hold registered with `PendingTaskID`. Yes/no prompt substituted into the model's tool_use_result.
5. User replies *approve* in chat → `ApproveInlineTask` CAS-transitions `pending_approval → active`, mints credential placeholders, resolves the canonical approval record.
6. User replies *deny* → `DenyInlineTask` transitions `pending_approval → denied`.
7. Abandoned chat-bound tasks past the 24h cache TTL get auto-denied by a new sweep so they don't sit forever in the dashboard.

## Dashboard guards

`TasksHandler.Approve` / `Deny` (HTTP) and `ApproveByTaskID` / `DenyByTaskID` (Telegram notifier callback) refuse chat-bound pending rows with `409 INLINE_CHAT_BOUND`. The chat surface bypasses via the dedicated `ApproveInlineTask` / `DenyInlineTask` methods — those skip the guard because they are the legitimate caller. `TaskCard` renders a *"Reply approve or deny in the agent chat"* notice in place of buttons when `approval_source === "inline_chat" && status === "pending_approval"`.

## Backwards compatibility

Legacy `InlineTaskCreator` implementations that don't expose the new `InlineTaskPendingCreator` extension still work — the intercept skips the pending-task landing and `resolveInlineTaskApproval` falls back to the create-on-approve flow via `resolved.TaskDefinition`. Existing tests cover that fallback path.

## Edge-case fixes (`bcc110bb`)

- `ApproveInlineTask`: CAS pending → active *before* minting credentials so a lost race to the expiry sweeper (which targets `approval_source='inline_chat'` rows) can't strand placeholders bound to a now-denied task. Mint failure post-CAS rolls back via `context.WithoutCancel`.
- `DenyInlineTask`: idempotent on already-terminal states (`denied` / `expired` / `revoked`) and lost CAS — the user gets the same denied reply either way, no spurious error text in `out.Reason`.
- Intercept failure audit decisions use `"fallthrough"` (not `"block"`) to match the file's audit-decision contract for `(verdict{}, false)` returns.
- Cache-hold-failed rollback uses `context.WithoutCancel(req.Context())` so a mid-request client disconnect doesn't strand an orphan pending task for the 24h TTL.

## Test plan

- [x] `go test ./internal/runtime/llmproxy/... ./internal/api/handlers/...` green.
- [x] Full `go test ./...` green except `TestLiteProxyScenarios/cross_file_inspection/codex` — model-behavior flakiness the scenario YAML itself is designed to catch (the intercept fires *after* the model decides to POST, so my changes can't affect what the model emits).
- [x] TypeScript build clean.
- [ ] Manual: confirm chat-bound pending tasks show on the Tasks page with the *"reply in agent chat"* notice and no Approve/Deny buttons.
- [ ] Manual: confirm a chat *approve* reply transitions the row to active and surfaces credentials to the model.
- [ ] Manual: confirm the dashboard's Approve endpoint returns 409 `INLINE_CHAT_BOUND` for a chat-bound pending row.
- [ ] Manual: confirm an abandoned chat-bound row (24h after creation, no reply) gets auto-denied by the sweeper.

### DIFF
--- a/internal/api/handlers/approvals.go
+++ b/internal/api/handlers/approvals.go
@@ -714,8 +714,14 @@ func (h *ApprovalsHandler) expireTimedOut(ctx context.Context) {
 	cutoff := time.Now().UTC().Add(-llmproxy.InlineTaskApprovalHoldTTL)
 	abandoned, err := h.st.ListExpiredInlineChatPendingTasks(ctx, cutoff)
 	if err != nil {
+		// Log and skip THIS block, but don't return — any sweep
+		// added below us must still run. Today the inline-chat
+		// sweep happens to be the terminal step in
+		// expireTimedOut, but returning here would silently
+		// shadow any future sweep added under this one if the
+		// list query started failing intermittently.
 		h.logger.WarnContext(ctx, "inline-chat pending expiry sweep: list failed", "err", err)
-		return
+		abandoned = nil
 	}
 	for _, task := range abandoned {
 		// Mark as "expired" (not "denied") to stay consistent with
--- a/internal/api/handlers/approvals.go
+++ b/internal/api/handlers/approvals.go
@@ -718,7 +718,11 @@ func (h *ApprovalsHandler) expireTimedOut(ctx context.Context) {
 		return
 	}
 	for _, task := range abandoned {
-		won, statusErr := h.st.UpdateTaskStatusFrom(ctx, task.ID, "pending_approval", "denied")
+		// Mark as "expired" (not "denied") to stay consistent with
+		// the sibling active-session expiry loop above and to keep
+		// the distinction between user-driven denial and timeout in
+		// the dashboard's Tasks page.
+		won, statusErr := h.st.UpdateTaskStatusFrom(ctx, task.ID, "pending_approval", "expired")
 		if statusErr != nil {
 			h.logger.WarnContext(ctx, "inline-chat pending expiry sweep: status flip failed",
 				"task_id", task.ID, "err", statusErr)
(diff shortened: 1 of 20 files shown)


ericlevine:
## Operator-facing audit-key change (heads up)

PR-review surfaced that two intercept failure paths reclassified from `audit(decision="block")` to `audit(decision="fallthrough")` — specifically `inline_task_pending_create_failed` and `inline_task_hold_failed`. The rationale (correct decision for a `(verdict{}, false)` return that falls through to the dashboard rewrite) is captured in the file's audit-decision comment at lines 99-106.

**Downstream alerting/dashboards keyed on `decision="block" AND outcome="inline_task_*"` will silently stop firing on these two outcomes** — they're now `decision="fallthrough"`. Re-key any operator dashboards before merging.

## Deferred follow-ups from review

- Add a covering index on `tasks(status, approval_source, created_at)` (postgres partial on `pending_approval`) once inline-chat traffic warrants — out of scope for this PR.
- Consider a small grace buffer past `InlineTaskApprovalHoldTTL` before the sweeper engages so a user replying right at 24h doesn't ambiguously race the sweep (correctness preserved either way via CAS).
- Add a `GetPendingApprovalRecordByTaskID` to avoid the O(N×M) scan in `resolvePendingTaskCanonicalRecord`.

PR #481, opened 2026-05-29
fix(llmproxy): sanitize user-supplied fields in task approval prompt

adityajha2005:
Strips ASCII control characters and Unicode directional overrides from user-supplied purpose and .Why fields before they're rendered into the inline task approval prompt.

### DIFF
--- a/internal/runtime/llmproxy/inline_task_prompt.go
+++ b/internal/runtime/llmproxy/inline_task_prompt.go
@@ -52,7 +52,7 @@ func renderTaskApprovalPromptWithRisk(req *runtimetasks.TaskCreateRequest, appro
 	if req == nil {
 		return "Clawvisor wants to create a task.\n\nReply `yes` or `y` to authorize, `no` or `n` to cancel." + suffix
 	}
-	purpose := strings.TrimSpace(req.Purpose)
+	purpose := sanitizeUserText(strings.TrimSpace(req.Purpose))
 	if purpose == "" {
 		return "Clawvisor wants to create a task: unnamed.\n\nReply `yes` or `y` to authorize, `no` or `n` to cancel." + suffix
 	}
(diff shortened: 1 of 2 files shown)


ericlevine:
Thanks for sending this in — really appreciate the well-scoped fix and the thorough test coverage (both the unit table and the end-to-end render assertion). Going to merge as-is.

A couple of follow-ups worth a second PR if you're up for it:

1. **Extend coverage to the other agent-supplied identifiers rendered into the same prompt.** `tool.ToolName`, `eg.Host`, and `cred.VaultItemID` / `cred.VaultItemHandle` are written into the bullet lines right next to the `.Why` fields and currently bypass `sanitizeUserText`. An attacker who can set `Why` can also set `ToolName`, so e.g. `ToolName: "Bash\u202eevil"` lands the RLO override back into the prompt and defeats the fix. Probably cleanest to sanitize at the write site rather than per-field so future additions don't have to remember.

2. **Consider broadening the blocklist to all Unicode format characters.** The current set covers bidi overrides but not ZWSP/ZWNJ/ZWJ (U+200B–D), BOM (U+FEFF), or soft hyphen (U+00AD), which are commonly used for invisible prompt-smuggling. Switching to `unicode.IsControl(r) || unicode.Is(unicode.Cf, r)` (while still preserving `\t`/`\n`/`\r`) catches the full category and won't bit-rot as new Unicode revisions add format chars.

Thanks again!

adityajha2005:
> Thanks for sending this in — really appreciate the well-scoped fix and the thorough test coverage (both the unit table and the end-to-end render assertion). Going to merge as-is.
> 
> A couple of follow-ups worth a second PR if you're up for it:
> 
> 1. **Extend coverage to the other agent-supplied identifiers rendered into the same prompt.** `tool.ToolName`, `eg.Host`, and `cred.VaultItemID` / `cred.VaultItemHandle` are written into the bullet lines right next to the `.Why` fields and currently bypass `sanitizeUserText`. An attacker who can set `Why` can also set `ToolName`, so e.g. `ToolName: "Bash\u202eevil"` lands the RLO override back into the prompt and defeats the fix. Probably cleanest to sanitize at the write site rather than per-field so future additions don't have to remember.
> 2. **Consider broadening the blocklist to all Unicode format characters.** The current set covers bidi overrides but not ZWSP/ZWNJ/ZWJ (U+200B–D), BOM (U+FEFF), or soft hyphen (U+00AD), which are commonly used for invisible prompt-smuggling. Switching to `unicode.IsControl(r) || unicode.Is(unicode.Cf, r)` (while still preserving `\t`/`\n`/`\r`) catches the full category and won't bit-rot as new Unicode revisions add format chars.
> 
> Thanks again!

sounds good, sending a follow up PR soon

PR #482, opened 2026-05-30
feat(lite-proxy): selectively stream SSE responses while preserving tool mediation

devcool20:
## Summary

This PR adds selective SSE streaming to Clawvisor’s lite-proxy response path.

Before this change, Clawvisor buffered the full upstream streaming response before writing anything back to the agent harness. That preserved tool-use safety, but it caused interactive terminals to appear frozen during long model responses. Users would wait for the full upstream response to complete, then receive the output all at once.

This PR changes the streaming boundary:

- assistant prose/text is forwarded immediately
- tool/function calls are buffered
- buffered tools are inspected by Clawvisor before reaching the harness
- blocked tools are withheld and replaced with the existing inline approval prompt
- allowed tools are re-emitted after policy evaluation

This keeps Clawvisor’s approval, credential, task-scope, and audit guarantees while restoring low-latency streaming for interactive agent harnesses.

## Why

The lite-proxy sits between agent harnesses and LLM providers. It needs to inspect tool calls before the agent can execute them, but it does not need to delay ordinary assistant text.

The previous full-buffering behavior made sense as a first safety implementation, but it created a poor experience for terminal-based agents. Long responses could take many seconds before the user saw any output.

Selective streaming solves that by treating content types differently:

- text is safe to forward as it arrives
- tool calls are action-bearing and must remain gated

### Selective SSE postprocessing

Adds a streaming postprocess path for SSE responses.

The new path forwards safe stream events immediately and buffers action-bearing tool events until Clawvisor can evaluate them.

This preserves the existing security model:

- tool calls are still inspected before delivery
- task-scope enforcement still applies
- credential mediation still applies
- approval-required tools are still withheld
- inline approval prompts still use Clawvisor’s existing flow
- audit/outcome tracking still records the request result

### Anthropic Messages streaming

Adds selective streaming support for Anthropic Messages-style SSE streams.

Covered behavior:

- forwards text blocks immediately
- preserves text-only stream terminal events
- buffers `tool_use` blocks
- preserves content block indexes when delayed tool blocks are re-emitted
- avoids duplicate/skipped content indexes that strict harness parsers reject
- emits approval prompts at a safe next content index when a tool is blocked
- keeps text-only streams free of synthetic tool stops

### OpenAI Chat Completions streaming

Adds selective handling for OpenAI Chat Completions-style streams.

Covered behavior:

- preserves text-only chat streaming
- preserves normal `finish_reason: "stop"`
- preserves `data: [DONE]`
- buffers tool-call chunks for post-stream evaluation
- re-emits buffered tool calls in stable index order after policy evaluation

### OpenAI Responses handling

Keeps the Responses API path conservative.

This PR avoids emitting Chat Completions-style tool chunks for Responses API streams. Text-only pass-through behavior remains covered, while Responses-specific tool re-emission can be expanded separately once fully modeled and tested.

### Synthetic history cleanup

Improves follow-up request cleanup after inline approval flows.

Clawvisor-generated approval prompts are stripped before forwarding later conversation history upstream. When stripping creates adjacent same-role messages, the history cleanup merges those messages so strict upstream protocol validators do not reject the next request.

This matters for harnesses that replay full conversation history after a tool approval or denial.

## Harness Coverage

This feature is harness-general. It is not specific to Claude Code.

The implementation sits at Clawvisor’s lite-proxy protocol boundary, so it applies to any harness that uses the supported Anthropic/OpenAI-compatible endpoints.

### Claude Code

Claude Code uses the Anthropic Messages-style path:

```text
ANTHROPIC_BASE_URL=<clawvisor>/api
POST /api/v1/messages
Claude Code benefits from immediate assistant text streaming while Clawvisor still buffers and mediates tool_use blocks.

Hermes
Hermes is covered when configured to use Clawvisor as its Anthropic-compatible or OpenAI-compatible model endpoint.

Expected paths:

POST /api/v1/messages
POST /api/v1/chat/completions
POST /api/v1/responses
For Hermes-style local agent workflows, this means long assistant text can stream through Clawvisor without waiting for the entire upstream response, while tool execution remains gated by Clawvisor’s task and approval policies.

OpenClaw
OpenClaw is covered under the same protocol-level contract.

If OpenClaw points its model traffic at Clawvisor’s lite-proxy endpoint, Clawvisor can selectively stream text while preserving mediation for tool/function calls.

Supported paths depend on the configured provider dialect:

Anthropic-compatible: POST /api/v1/messages
OpenAI Chat-compatible: POST /api/v1/chat/completions
OpenAI Responses-compatible: POST /api/v1/responses
The feature is therefore reusable across OpenClaw deployments that use Clawvisor as the model gateway.

Codex / OpenAI-compatible Harnesses
OpenAI-compatible harnesses are covered through:

OPENAI_BASE_URL=<clawvisor>/api/v1
POST /api/v1/chat/completions
POST /api/v1/responses
Chat Completions streaming receives the same selective behavior: text deltas stream immediately, while tool calls remain buffered for inspection.

Windows Note
Manual validation on Windows has one important caveat.

Some terminal harnesses may not visually render line-by-line output on Windows even when SSE bytes are forwarded correctly underneath. In that case, the reliable proof is the Clawvisor log timing, not the terminal paint behavior.

For this PR, the important evidence is:

upstream SSE headers are received
Clawvisor forwards first SSE bytes early
Clawvisor forwards the first text delta early
the request completes later
That proves Clawvisor is no longer waiting for full upstream completion before writing client-facing stream data.

Manual Verification
Local gateway readiness:

Invoke-RestMethod -Uri http://localhost:25297/ready
Observed result:

db status vault
-- ------ -----
ok ok     ok
Focused mock SSE timing proof:

go test ./internal/api/handlers -run TestLLMEndpoint_SSETextStreamsBeforeUpstreamCompletes -count=1 -v
Observed log evidence:

=== RUN   TestLLMEndpoint_SSETextStreamsBeforeUpstreamCompletes
2026/05/30 19:20:33 INFO lite-proxy upstream headers received request_id=300c6b0a-25e9-4fe9-8970-78ba2b423229 agent_id=1fa647cb-2ad4-464d-8921-217cf4b1b060 provider=anthropic upstream_url=http://127.0.0.1:58598/v1/messages status=200 content_type=text/event-stream anthropic_request_id="" openai_request_id="" ttfb_headers_ms=1
2026/05/30 19:20:33 INFO lite-proxy stream forwarded first SSE bytes request_id=300c6b0a-25e9-4fe9-8970-78ba2b423229 provider=anthropic elapsed_ms=0 ttfb_headers_ms=1 chunk_bytes=21
2026/05/30 19:20:33 INFO lite-proxy stream forwarded text block start request_id=300c6b0a-25e9-4fe9-8970-78ba2b423229 provider=anthropic elapsed_ms=0 ttfb_headers_ms=1
2026/05/30 19:20:33 INFO lite-proxy stream forwarded first text delta request_id=300c6b0a-25e9-4fe9-8970-78ba2b423229 provider=anthropic elapsed_ms=1 ttfb_headers_ms=1
2026/05/30 19:20:35 INFO lite-proxy request completed request_id=300c6b0a-25e9-4fe9-8970-78ba2b423229 agent_id=1fa647cb-2ad4-464d-8921-217cf4b1b060 action=lite_proxy.messages.create http_status=200 decision=allow outcome=success reason="" caller_auth_source=authorization client_cancelled=false total_ms=1506
--- PASS: TestLLMEndpoint_SSETextStreamsBeforeUpstreamCompletes (1.84s)
PASS
ok github.com/clawvisor/clawvisor/internal/api/handlers 2.110s
This proves the core timing guarantee: first SSE bytes and first text delta are forwarded before upstream completion.

Test Coverage
Automated checks run locally:

go test ./internal/api/handlers -run "TestValidateLLMAPIKey|Test.*Stream|Test.*Streaming|TestLLMEndpoint" -count=1
Result:

ok github.com/clawvisor/clawvisor/internal/api/handlers 12.414s
go test ./internal/runtime/conversation/... -count=1
Result:

ok github.com/clawvisor/clawvisor/internal/runtime/conversation 1.720s
go test ./internal/runtime/llmproxy/... -run "Test.*Stream|Test.*Streaming|TestSynthetic|TestMerge|TestContinuation|TestSuppressOpenRouterGPTOSSReasoning" -count=1
Result:

ok github.com/clawvisor/clawvisor/internal/runtime/llmproxy 0.822s
ok github.com/clawvisor/clawvisor/internal/runtime/llmproxy/inspector 2.657s [no tests to run]
ok github.com/clawvisor/clawvisor/internal/runtime/llmproxy/pricing 1.227s [no tests to run]
Focused SSE timing check:

go test ./internal/api/handlers -run TestLLMEndpoint_SSETextStreamsBeforeUpstreamCompletes -count=1 -v
Result:

PASS
ok github.com/clawvisor/clawvisor/internal/api/handlers 2.110s


ericlevine:
Thanks for taking this on — the timing proof in the test logs is exactly the kind of evidence I want to see, and the index-preservation work for the Anthropic stream is careful. A few things I'd want addressed before this lands.

**Blockers**

1. **Continuation is silently dropped in the streaming path.** `PostprocessStream` populates `ContinuationToolResults`, `AssistantTurn`, `StreamingProvider`, and `StreamingResult` on the result, but the streaming branch in `llm_endpoint.go` (around line 893, the `return` at the end of the `if conversation.IsSSEContentType(...)` block) never calls `tryContinuation`. The non-streaming path does. So any verdict with `ContinueWithToolResult` becomes a one-shot fallback-text emission with no model continuation — the model never sees the synthetic tool_result. This is a behavioral regression versus the JSON path and contradicts the "existing security model is preserved" claim in the description. Either wire continuation into the streaming branch or explicitly call out streaming as continuation-incompatible in the PR description and add a test that asserts the current behavior.

2. **`custom_tool_call` bypasses inspection in `streamRewriteResponses`** (`openai_response.go:1849`). Custom tool calls get appended to `orderedItems` with `isCustomToolCall: true` and rendered as assistant fragments, but they never enter `streamResult.ToolUses`, so `PostprocessStream`'s eval loop skips them entirely. That's an inspector bypass on the Responses stream. Please add a test case that sends a `custom_tool_call` containing a credentialed call and asserts it's inspected/blocked, then route custom tool calls through the same eval path as regular function calls.

3. **Possible double `response.completed` for OpenAI Responses.** `writeProviderStop` for `openai_responses` is a no-op, but `writeProviderToolUses` → `SynthOpenAIResponsesFunctionCallsSSE` emits its own lifecycle envelope. Depending on which upstream events made it through before the rewrite kicked in, harnesses can see two completion events. Worth a test that asserts exactly one `response.completed` in the rewritten output.

**Worth fixing before merge**

4. **Duplicated authorization pipeline.** `innerEval` in `PostprocessStream` reimplements the verdict pipeline from `Postprocess` almost line-for-line — control parse, inspector inspect, decision evaluate, scope check, nonce mint, rewrite. Future security changes will have to be made in two places, and the drift between them is the kind of thing that becomes an exploit. Can the shared logic be factored into a helper that both paths call?

5. **Hold storage path runs twice.** `innerEval` calls `cfg.PendingApprovals.Hold(...)` via the `holdCapturingApprovalCache` wrapper, and then `replayBufferedHolds` runs against the original cache after. I haven't been able to convince myself that evictions and audit rows aren't being emitted twice — could you walk through that flow or add a test that pins down the call count?

6. **Dead code.** `WriteStreamingContinuationFallback` is a `return nil` with a test that asserts it's a no-op and no production callers. Please remove.

7. **Unused flag.** `AllowAnthropicCompatModelPatch` is set only in tests — no production caller passes `true`. Either wire it up or drop the flag and the unused code path.

**Minor**

- `stripAnthropicCacheControl` round-trips through `map[string]any`, which reorders JSON keys on every flagged body — works but worth a comment noting it.
- `isNonClaudeModel` uses `strings.Contains(model, "claude")`, so `myclaude-clone` would skip filtering.
- `streamingTimingWriter` matches `"type":"text"` via `bytes.Contains`, which will false-fire on tool inputs that happen to contain that substring. Observational only.
- Three different `writeSSE` functions across the conversation and llmproxy packages with different signatures — would help future readers to consolidate or rename.

Happy to dig into any of these in more detail.

devcool20:
> Thanks for taking this on — the timing proof in the test logs is exactly the kind of evidence I want to see, and the index-preservation work for the Anthropic stream is careful. A few things I'd want addressed before this lands.
> 
> **Blockers**
> 
> 1. **Continuation is silently dropped in the streaming path.** `PostprocessStream` populates `ContinuationToolResults`, `AssistantTurn`, `StreamingProvider`, and `StreamingResult` on the result, but the streaming branch in `llm_endpoint.go` (around line 893, the `return` at the end of the `if conversation.IsSSEContentType(...)` block) never calls `tryContinuation`. The non-streaming path does. So any verdict with `ContinueWithToolResult` becomes a one-shot fallback-text emission with no model continuation — the model never sees the synthetic tool_result. This is a behavioral regression versus the JSON path and contradicts the "existing security model is preserved" claim in the description. Either wire continuation into the streaming branch or explicitly call out streaming as continuation-incompatible in the PR description and add a test that asserts the current behavior.
> 2. **`custom_tool_call` bypasses inspection in `streamRewriteResponses`** (`openai_response.go:1849`). Custom tool calls get appended to `orderedItems` with `isCustomToolCall: true` and rendered as assistant fragments, but they never enter `streamResult.ToolUses`, so `PostprocessStream`'s eval loop skips them entirely. That's an inspector bypass on the Responses stream. Please add a test case that sends a `custom_tool_call` containing a credentialed call and asserts it's inspected/blocked, then route custom tool calls through the same eval path as regular function calls.
> 3. **Possible double `response.completed` for OpenAI Responses.** `writeProviderStop` for `openai_responses` is a no-op, but `writeProviderToolUses` → `SynthOpenAIResponsesFunctionCallsSSE` emits its own lifecycle envelope. Depending on which upstream events made it through before the rewrite kicked in, harnesses can see two completion events. Worth a test that asserts exactly one `response.completed` in the rewritten output.
> 
> **Worth fixing before merge**
> 
> 4. **Duplicated authorization pipeline.** `innerEval` in `PostprocessStream` reimplements the verdict pipeline from `Postprocess` almost line-for-line — control parse, inspector inspect, decision evaluate, scope check, nonce mint, rewrite. Future security changes will have to be made in two places, and the drift between them is the kind of thing that becomes an exploit. Can the shared logic be factored into a helper that both paths call?
> 5. **Hold storage path runs twice.** `innerEval` calls `cfg.PendingApprovals.Hold(...)` via the `holdCapturingApprovalCache` wrapper, and then `replayBufferedHolds` runs against the original cache after. I haven't been able to convince myself that evictions and audit rows aren't being emitted twice — could you walk through that flow or add a test that pins down the call count?
> 6. **Dead code.** `WriteStreamingContinuationFallback` is a `return nil` with a test that asserts it's a no-op and no production callers. Please remove.
> 7. **Unused flag.** `AllowAnthropicCompatModelPatch` is set only in tests — no production caller passes `true`. Either wire it up or drop the flag and the unused code path.
> 
> **Minor**
> 
> * `stripAnthropicCacheControl` round-trips through `map[string]any`, which reorders JSON keys on every flagged body — works but worth a comment noting it.
> * `isNonClaudeModel` uses `strings.Contains(model, "claude")`, so `myclaude-clone` would skip filtering.
> * `streamingTimingWriter` matches `"type":"text"` via `bytes.Contains`, which will false-fire on tool inputs that happen to contain that substring. Observational only.
> * Three different `writeSSE` functions across the conversation and llmproxy packages with different signatures — would help future readers to consolidate or rename.
> 
> Happy to dig into any of these in more detail.

Thanks for the detailed review @ericlevine. I addressed the blockers and hardening items in this pass.

Continuation is now wired for the streaming path instead of being reduced to fallback text. `PostprocessStream` returns structured continuation state, and the lite-proxy handler invokes the continuation request with the captured assistant turn plus synthetic `tool_result`.

I also fixed the provider wire-shape issue you called out: streaming continuations are now spliced into the already-open SSE envelope instead of appending a second complete provider stream. Anthropic continuations drop the second `message_start`, offset continuation content indexes, and preserve one final `message_stop`. OpenAI Responses continuations drop the second `response.created`, offset continuation `output_index`, keep the original response id on completion, and emit exactly one `response.completed`.

For Responses custom tool calls, streamed `custom_tool_call` items are now withheld from passthrough, converted into `ToolUse`, routed through the same inspection/blocking path, and covered by a regression test that asserts the credentialed call is blocked and does not leak before inspection.

I also added coverage for hold capture/replay committing each hold exactly once, removed the dead streaming continuation fallback helper, and removed the local Anthropic/OpenRouter compatibility patch flag/path so synthetic history stripping only handles Clawvisor-generated approval history.

All validation tests pass perfectly.

ericlevine:
Thanks for the quick turnaround on this — the new commit addresses all three blockers I'd raised, and I appreciate the test coverage you added alongside each fix.

What I confirmed on the latest:

- Continuation is wired into the streaming path, the splice helpers correctly drop the second `message_start` / `response.created` and offset content indices, and `TestTryContinuation_PostsSecondCallWithSSEToolResult` plus the two splice tests pin down the behavior.
- `custom_tool_call` is now flagged via `withheldTool` on both the `output_item.added` and `output_item.done` paths, threaded into `streamResult.ToolUses`, and inspected by `PostprocessStream`. The new `TestPostprocessStream_OpenAIResponsesCustomToolCallIsInspectedAndBlocked` is exactly the regression test I was hoping for.
- The `!withheldTool` gate replaces `len(pending) == 0`, which closes the double-`response.completed` gap, and the count assertion in the test makes that explicit.
- `WriteStreamingContinuationFallback`, `streamingContinuationFallbackText`, and `AllowAnthropicCompatModelPatch` (plus the cache_control / thinking strip code path) are all removed — that drops the unrelated scope cleanly.
- `isNonClaudeModel` now strips the provider prefix and checks `claude-` as a prefix, and `streamingTimingWriter` parses SSE events rather than substring-matching, so the false-positive paths are gone.
- `TestHoldCapturingApprovalCacheBuffersThenReplayCommitsOnce` pins the buffering-then-replay invariant I was worried about. That's the concrete property I wanted asserted — thanks for adding it.

A few non-blocking items I'd still flag:

1. **Authorization pipeline duplication remains.** `innerEval` in `PostprocessStream` still re-implements the verdict pipeline from `Postprocess` (control parse → inspect → decision evaluate → scope → nonce → rewrite) almost line-for-line. I don't think it has to block this PR, but it's the kind of drift that turns into a security bug the first time someone fixes one branch and forgets the other. Could we open a follow-up to factor the shared pipeline into a helper that both paths call?

2. **`mergeAuditUsage` parameter order.** `(..., logger *slog.Logger, ctx context.Context, ...)` reads against the Go convention of ctx first. Trivial nit.

3. **Splice helpers parse SSE via `strings.Split("\n\n")`.** Safe here because they only consume bodies produced by our own rewriter (which JSON-encodes data lines, so embedded `\n\n` can't happen), but it would be worth a one-line comment on `spliceStreamingContinuationBody` noting that precondition so a future caller doesn't reach for it on arbitrary upstream SSE.

Let me know if you want to address these or address in a follow-up. Once we're set on the code quality side, I'll test this all locally to make sure it's working as expected with a few agent harnesses.

devcool20:
> Thanks for the quick turnaround on this — the new commit addresses all three blockers I'd raised, and I appreciate the test coverage you added alongside each fix.
> 
> What I confirmed on the latest:
> 
> * Continuation is wired into the streaming path, the splice helpers correctly drop the second `message_start` / `response.created` and offset content indices, and `TestTryContinuation_PostsSecondCallWithSSEToolResult` plus the two splice tests pin down the behavior.
> * `custom_tool_call` is now flagged via `withheldTool` on both the `output_item.added` and `output_item.done` paths, threaded into `streamResult.ToolUses`, and inspected by `PostprocessStream`. The new `TestPostprocessStream_OpenAIResponsesCustomToolCallIsInspectedAndBlocked` is exactly the regression test I was hoping for.
> * The `!withheldTool` gate replaces `len(pending) == 0`, which closes the double-`response.completed` gap, and the count assertion in the test makes that explicit.
> * `WriteStreamingContinuationFallback`, `streamingContinuationFallbackText`, and `AllowAnthropicCompatModelPatch` (plus the cache_control / thinking strip code path) are all removed — that drops the unrelated scope cleanly.
> * `isNonClaudeModel` now strips the provider prefix and checks `claude-` as a prefix, and `streamingTimingWriter` parses SSE events rather than substring-matching, so the false-positive paths are gone.
> * `TestHoldCapturingApprovalCacheBuffersThenReplayCommitsOnce` pins the buffering-then-replay invariant I was worried about. That's the concrete property I wanted asserted — thanks for adding it.
> 
> A few non-blocking items I'd still flag:
> 
> 1. **Authorization pipeline duplication remains.** `innerEval` in `PostprocessStream` still re-implements the verdict pipeline from `Postprocess` (control parse → inspect → decision evaluate → scope → nonce → rewrite) almost line-for-line. I don't think it has to block this PR, but it's the kind of drift that turns into a security bug the first time someone fixes one branch and forgets the other. Could we open a follow-up to factor the shared pipeline into a helper that both paths call?
> 2. **`mergeAuditUsage` parameter order.** `(..., logger *slog.Logger, ctx context.Context, ...)` reads against the Go convention of ctx first. Trivial nit.
> 3. **Splice helpers parse SSE via `strings.Split("\n\n")`.** Safe here because they only consume bodies produced by our own rewriter (which JSON-encodes data lines, so embedded `\n\n` can't happen), but it would be worth a one-line comment on `spliceStreamingContinuationBody` noting that precondition so a future caller doesn't reach for it on arbitrary upstream SSE.
> 
> Let me know if you want to address these or address in a follow-up. Once we're set on the code quality side, I'll test this all locally to make sure it's working as expected with a few agent harnesses.

Addressed the three follow-up items:

- Factored the duplicated JSON/SSE tool-use verdict pipeline into a shared helper used by both Postprocess and PostprocessStream.
- Reordered mergeAuditUsage args so context comes before logger.
- Added the spliceStreamingContinuationBody precondition comment for Clawvisor-generated SSE.

ericlevine:
Code looks great, tested on different harnesses, everything seems to be in order. Merged! Thanks @devcool20

PR #502, opened 2026-06-02
fix(llmproxy): pass preferred task id during approval release

devcool20:
### Description
When running multiple agents on the same token in parallel, task checkout/focus can shift. During the approval release re-check (`TryReleasePendingApproval`), the evaluation input was not passing the originally matched task context (`PreferredTaskID`). This caused Clawvisor to evaluate the check in a vacuum or under a shifted focus, resulting in a fingerprint mismatch and throwing the error:
> `⏺ Clawvisor: couldn't release this approval. held approval no longer matches current authorization decision`

This PR fixes this by ensuring the originally matched task ID (`held.Fingerprint.TaskID`) is correctly restored as the `PreferredTaskID` during the recheck evaluation.

### Changes
* **`internal/runtime/llmproxy/release.go`**: Set `PreferredTaskID` to `held.Fingerprint.TaskID` when reconstructing the `AuthorizationInput` for both trigger-miss and credentialed rechecks.
* **`internal/runtime/llmproxy/release_test.go`**: Added `TestTryReleasePendingApproval_ParallelPreferredTaskID` to verify that release rechecks succeed under parallel task checkout shifts.

### Verification
* Ran unit tests locally: `go test -v ./internal/runtime/llmproxy/...` (Passed).
* Manually verified in a local Docker Compose setup by triggering two parallel agent task approvals and confirming they release successfully without conflict.

### DIFF
--- a/internal/runtime/llmproxy/release.go
+++ b/internal/runtime/llmproxy/release.go
@@ -327,8 +327,8 @@ func rewriteApprovedHeldToolUse(ctx context.Context, req ReleaseRequest, held He
 			CandidateTasks:    req.CandidateTasks,
 			ToolRules:         req.ToolRules,
 			EgressRules:       req.EgressRules,
-			PreferredTaskID:   held.Fingerprint.TaskID,
 			IntentVerifier:    decisionIntentVerifier{inner: req.IntentVerifier},
+			PreferredTaskID:   held.Fingerprint.TaskID,
 			AllowMissingScope: true,
 		}
 		dec, err := runtimedecision.EvaluateAuthorization(ctx, decisionInput)
@@ -373,8 +373,8 @@ func rewriteApprovedHeldToolUse(ctx context.Context, req ReleaseRequest, held He
 		CandidateTasks:  req.CandidateTasks,
 		ToolRules:       req.ToolRules,
 		EgressRules:     req.EgressRules,
-		PreferredTaskID: held.Fingerprint.TaskID,
 		IntentVerifier:  decisionIntentVerifier{inner: req.IntentVerifier},
+		PreferredTaskID: held.Fingerprint.TaskID,
 	}
 	dec, err := runtimedecision.EvaluateAuthorization(ctx, decisionInput)
 	if err != nil {
--- a/internal/runtime/llmproxy/release.go
+++ b/internal/runtime/llmproxy/release.go
@@ -327,6 +327,7 @@ func rewriteApprovedHeldToolUse(ctx context.Context, req ReleaseRequest, held He
 			CandidateTasks:    req.CandidateTasks,
 			ToolRules:         req.ToolRules,
 			EgressRules:       req.EgressRules,
+			PreferredTaskID:   held.Fingerprint.TaskID,
 			IntentVerifier:    decisionIntentVerifier{inner: req.IntentVerifier},
 			AllowMissingScope: true,
 		}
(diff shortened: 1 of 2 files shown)


ericlevine:
Thanks for tracking this down — the production fix is on the money. Restoring `PreferredTaskID` from `held.Fingerprint.TaskID` is the right move: it replays the original decision context instead of the current (possibly shifted) focus, and since `PreferredTaskID` only tie-breaks among already-matching tasks (see the doc on `AuthorizationInput.PreferredTaskID` in `pkg/runtime/decision/decision.go`), there's no scope-widening risk. Both `AuthorizationInput{}` sites in `release.go` are correctly patched.

One important issue with the test, plus a couple of smaller items:

**`TestTryReleasePendingApproval_ParallelPreferredTaskID` doesn't actually exercise the bug.**

I checked out the branch, reverted `release.go` to `origin/main`, ran `go clean -testcache`, and re-ran the test — it still passes. The reason: `taskA` and `taskB` have identical `ExpectedTools` scope, and `MatchToolCallPreferred`'s fallback in `internal/runtime/policy/task_match.go` uses strict `score > bestScore`, so the first task in slice order wins on ties. At both hold and release, candidates are `[taskA, taskB]`, so the fallback picks `task-A` regardless of `PreferredTaskID` — fingerprint matches either way. The "focus shift" the PR description targets isn't modeled.

To make this a real regression test, you need an observable focus shift. Either:
- Reorder `CandidateTasks` at release time to `[taskB, taskA]` (simulates a checkout swap), or
- Give `taskB` a stricter `ExpectedTools` entry so it outscores `taskA` in the fallback.

Either makes the release-time fingerprint depend on `PreferredTaskID` being restored from the hold — which is the actual production bug. Without that, this test will keep passing even if someone deletes the fix later.

**Smaller items:**
- `gofmt -l internal/runtime/llmproxy/release_test.go` flags the file (looks like trailing whitespace around the `seedPostprocessStoreWithService` call line). A `gofmt -w` should clean it up.
- The test only covers the trigger-miss branch (no `Target`/`Service` set, `Catalog` nil). The credentialed-API branch at `release.go:365` got the same fix but no dedicated coverage — worth a parallel case where the inspector resolves to a real host/method/path so both code paths are guarded.

Happy to re-review once the test actually fails without the production change.

devcool20:
> Thanks for tracking this down — the production fix is on the money. Restoring `PreferredTaskID` from `held.Fingerprint.TaskID` is the right move: it replays the original decision context instead of the current (possibly shifted) focus, and since `PreferredTaskID` only tie-breaks among already-matching tasks (see the doc on `AuthorizationInput.PreferredTaskID` in `pkg/runtime/decision/decision.go`), there's no scope-widening risk. Both `AuthorizationInput{}` sites in `release.go` are correctly patched.
> 
> One important issue with the test, plus a couple of smaller items:
> 
> **`TestTryReleasePendingApproval_ParallelPreferredTaskID` doesn't actually exercise the bug.**
> 
> I checked out the branch, reverted `release.go` to `origin/main`, ran `go clean -testcache`, and re-ran the test — it still passes. The reason: `taskA` and `taskB` have identical `ExpectedTools` scope, and `MatchToolCallPreferred`'s fallback in `internal/runtime/policy/task_match.go` uses strict `score > bestScore`, so the first task in slice order wins on ties. At both hold and release, candidates are `[taskA, taskB]`, so the fallback picks `task-A` regardless of `PreferredTaskID` — fingerprint matches either way. The "focus shift" the PR description targets isn't modeled.
> 
> To make this a real regression test, you need an observable focus shift. Either:
> 
> * Reorder `CandidateTasks` at release time to `[taskB, taskA]` (simulates a checkout swap), or
> * Give `taskB` a stricter `ExpectedTools` entry so it outscores `taskA` in the fallback.
> 
> Either makes the release-time fingerprint depend on `PreferredTaskID` being restored from the hold — which is the actual production bug. Without that, this test will keep passing even if someone deletes the fix later.
> 
> **Smaller items:**
> 
> * `gofmt -l internal/runtime/llmproxy/release_test.go` flags the file (looks like trailing whitespace around the `seedPostprocessStoreWithService` call line). A `gofmt -w` should clean it up.
> * The test only covers the trigger-miss branch (no `Target`/`Service` set, `Catalog` nil). The credentialed-API branch at `release.go:365` got the same fix but no dedicated coverage — worth a parallel case where the inspector resolves to a real host/method/path so both code paths are guarded.
> 
> Happy to re-review once the test actually fails without the production change.

Thanks for the spot-on feedback, Eric!

I've pushed updates addressing all of your points:

Regression Proof: Swapped the candidate task list order to [taskB, taskA] at release time, and configured the mock IntentVerifier to refuse the intent (Allow: false). This forces a VerdictNeedsApproval decision, meaning reverting the release.go change now reliably triggers the fingerprint mismatch error and fails the tests.
Parallel Coverage: Added a dedicated test for the credentialed API release path (_Credentialed).
Format: Ran gofmt -w to clean up the trailing whitespace.

ericlevine:
Looks great! Thanks for addressing the feedback quickly. Merged.

PR #503, opened 2026-06-02
fix(llmproxy): decline tool_use rewrite under extended thinking

ericlevine:
## Summary

- When an Anthropic assistant message contains a \`thinking\` or \`redacted_thinking\` block, the model's signature covers every sibling block including each \`tool_use.input\`. The proxy's in-place rewrite of \`tool_use.input\` (to redirect credentialed curls at the resolver) silently invalidates the signature. The agent CLI echoes the rewritten content back to Anthropic on the next turn and Anthropic returns \`400 \\\`thinking\\\` or \\\`redacted_thinking\\\` blocks in the latest assistant message cannot be modified\`.
- Detect thinking/redacted-thinking presence at three points in the rewrite pipeline (\`rewriteJSON\`, \`rewriteSSE\`, and the streaming postprocess loop that applies rewrites after \`StreamRewrite\` returns) and convert any \`RewriteInput\`-bearing verdict into a structured refusal verdict. The downstream block-rendering path surfaces a Clawvisor refusal in place of the broken rewrite, so the conversation continues instead of hard-failing with an opaque 400.
- Three regression tests in \`response_test.go\` cover thinking, redacted-thinking, and the no-thinking baseline (to guard against the guard over-triggering).

## Who's affected

Any Anthropic-based agent with extended thinking enabled doing credentialed tool calls through the resolver — most prominently Claude Code on Sonnet 4.6+, which enables thinking on most reasoning-heavy task types. OpenAI/Codex paths aren't affected (no thinking-signature equivalent).

## Net behavior change

A turn that previously hard-failed with an opaque Anthropic 400 now fails closed with a Clawvisor-authored refusal on the same turn. The agent sees a tool_result explaining why the call was declined and the conversation continues.

## Test plan

- [x] \`go test ./internal/runtime/conversation/ ./internal/runtime/llmproxy/ ./internal/api/handlers/\` — all green.
- [x] New unit tests assert the JSON rewriter declines under \`thinking\` and \`redacted_thinking\`, and still admits rewrites on turns without either.
- [x] Manual e2e against Claude (Sonnet 4.6) with a credentialed Gmail-style request previously reproduced the upstream 400; with this fix the proxy now emits a Clawvisor refusal on the same turn instead.

### DIFF
--- a/internal/runtime/conversation/anthropic_response.go
+++ b/internal/runtime/conversation/anthropic_response.go
@@ -53,6 +53,25 @@ func (rw AnthropicResponseRewriter) rewriteJSON(body []byte, eval ToolUseEvaluat
 		return RewriteResult{Body: body}, nil
 	}
 
+	// Extended-thinking sentinel: when the assistant message has a
+	// thinking or redacted_thinking block, the model's signature
+	// covers every sibling block including tool_use.input. Rewriting
+	// the input in place would invalidate the signature, and the
+	// agent's NEXT request — which echoes this assistant message
+	// back to Anthropic with the rewritten input — would 400 with
+	// "thinking blocks in the latest assistant message cannot be
+	// modified." Detect the case here and convert any
+	// rewrite-bearing verdict into a structured block result so the
+	// downstream block-rendering path emits a Clawvisor refusal in
+	// place of the broken rewrite.
+	hasThinking := false
+	for _, block := range resp.Content {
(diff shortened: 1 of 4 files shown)


ericlevine:
Abandoning this in favor of continuing the script-sessions work. The bug is real but the fix path needs more design (rolling back the rewrite is a regression on Claude+thinking+credentialed workflows, not a true fix); separating the thinking-block work from the script-sessions PR keeps each PR focused. Will revisit as a dedicated workstream.

PR #506, opened 2026-06-02
feat(autovault): script sessions for credentialed fan-out workflows

ericlevine:
## Summary

Adds a first-class script-session path so agents can run credentialed multi-request workflows without Clawvisor rewriting arbitrary script shapes. Agent mints a short-lived, tightly-scoped session via \`POST /api/control/autovault/script-session\` and then issues N proxy calls under the returned caller token. The resolver enforces session bounds (host, methods, path prefixes, max_uses, TTL, per-request and aggregate byte caps) on every request and emits per-call audit rows.

Replaces the one-shot rewriter for credentialed fan-out (e.g., listing then fetching N Gmail message metadatas): one mint round-trip instead of N inspector/rewriter round-trips, with strictly tighter audit + scope semantics. Default caps: TTL ≤ 120s, max_uses ≤ 50, GET-only, single host + placeholder per session, 1 MiB per-request / 10 MiB aggregate.

## What's included

- **\`ScriptSessionCache\`** sibling to \`CallerNonceCache\` — multi-use scoped tokens with optimistic byte reservation in \`Authorize()\` and a \`ReleaseAuthorize()\` refund path that undoes both the reservation and \`UsedCount\` on deny-paths that don't reach the resolver.
- **Mint endpoint** with deterministic checks (placeholder ownership, bound-host allowlist, hard limits) + intent verification on the derived capability. Centralized \`denyMint\` helper emits a mint audit row on every rejection. 64 KiB body cap.
- **Discoverability**: \`GET /api/control/autovault/script\` static docs page; \`CREDENTIALED FAN-OUT\` section in the injected LLM control notice that names the trigger and the flow.
- **Resolver wiring**: accepts \`cv-script-*\` tokens, restricted to \`/api/proxy/*\` routes, enforces byte caps via a \`streamCap\` derived from session reservation + remaining aggregate budget, strips upstream \`Content-Length\` on truncation, releases the reservation via a defer that runs on every exit (happy, error, scope mismatch). Defense-in-depth check refuses requests whose swapped placeholder doesn't match the session's bound placeholder. Fails closed when the script-session cache isn't wired.
- **Inspector-passthrough gate**: when an agent emits a curl already shaped for the proxy (\`cv-script-\` in \`X-Clawvisor-Caller\`, URL pointing at the resolver), the inspector pipeline skips re-rewriting. Uses mvdan/sh to actually parse curl arguments — disqualifies URL-bearing / config-reading flags (\`--url\`, \`-K\`/\`--config\`, \`-L\`/\`--location\`, \`-L-trusted\`, \`-x\`/\`--proxy\`, \`--next\`, \`--connect-to\`, \`--resolve\`, \`--socks*\`, \`-T\`/\`--upload-file\`, \`--dns-servers\`) and combined-short-flag variants of those, closing off-proxy bypass surfaces (query-string smuggle, attacker URL with proxy-host substring, etc.).
- **Rewriter no-shape failure** redirects to the script-session mint endpoint with templated example parameters derived from the inspector verdict.
- **Audit**: \`LogScriptSessionMint\` and \`LogScriptSessionUse\` with proper \`TaskID\` columns, \`DedupKey\` shaped per row class (with UUID tiebreaker when \`X-Request-Id\` is missing), distinct outcome codes for "no bound service" vs "host out of allowlist," and \`script_session_total_cap_after_stream\` for post-stream cap breaches that don't flip the response \`decision\`.

## Tests

- **\`script_session_test.go\`**: 26-case table-driven test on the inspector-passthrough gate covering legitimate shapes (bash + structured), substring bypass attempts, off-proxy URLs (including \`?ref=…\` smuggle), config flags, combined-short variants (\`-Lk\`), mixed \`url\`+\`cmd\` shapes; plus cache-primitive coverage (Authorize/RecordBytes/ReleaseAuthorize/Revoke, optimistic reservation true-up, concurrent safety, TTL).
- **\`script_session_path.go\`**: prefix normalization, percent-encoded traversal rejection.
- **\`proxy_resolver_script_session_test.go\`**: happy path, scope mismatch, exhaustion, aggregate-cap-bounds-crossing-request, \`Content-Length\` stripping on truncation.
- **\`control_autovault_script_test.go\`**: mint validation, deterministic deny paths, deny-path audit emission.
- **Lite e2e**: harness wired with a counting \`ScriptSessionCache\` wrapper + two scenarios (\`script_session_credentialed_fanout\`, \`script_session_scope_mismatch_recovery\`) written in naturalistic user voice — they probe whether the agent infers from goal-shaped instructions rather than recall an exact prescriptive recipe.

## Review history

Iterated through ~10 rounds of \`cubic review\` against \`main\`. Final pass returns 0 issues at every priority (P0–P3). Known-and-documented limitations: Authorize-time denials don't emit a use-row (caller doesn't have agent context); the lite e2e harness's simplified resolver releases the byte reservation as zero rather than measuring real bytes (no scenario asserts on per-session byte totals).

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./internal/runtime/llmproxy/ ./internal/api/handlers/ ./internal/api/middleware/ ./internal/runtime/conversation/\` green
- [x] Lite e2e scenario YAML parses under strict decode (\`TestLoadAllLibraryScenarios\`)
- [x] Live Claude e2e: \`script_session_scope_mismatch_recovery/claude\` passes against real Anthropic API. \`script_session_credentialed_fanout/claude\` reaches the script-session path (\`script_session.mint: 1\`, \`downstream.placeholder_used: 1\`) but flakes on an unrelated Anthropic thinking-block issue; that's tracked as a separate problem (the agent's continued workflow remains gated on the model deciding to follow the prompt's CREDENTIALED FAN-OUT guidance, which is inconsistent today).


ericlevine:
@cubic-dev-ai review

PR #507, opened 2026-06-03
fix(llmproxy): preserve index 0 for thinking blocks in Anthropic SSE stream

devcool20:
### Description
Addresses the issue where Claude Code's "Percolating..." thinking display disappears prematurely when streaming responses through Clawvisor.

### Root Cause
Clawvisor prepended the first-turn notice block at Index 0 and pushed the model's actual streaming output to Index 1. Because Claude Code's parser expects any reasoning/thinking block to be at Index 0, shifting the thinking block to Index 1 caused the terminal harness to assume thinking was skipped or completed, instantly hiding the spinner.

### Solution
- Modified the Anthropic SSE stream rewriter in `streaming_assistant_prepend.go` to delay notice injection when a `thinking` block is detected first.
- The `thinking` block remains at Index 0, allowing client harnesses to render the reasoning spinner correctly.
- The notice is injected as a separate text block at Index 1 immediately before the model's output text block starts.

### Verification & Tests
- Added `TestStreamingFirstTurnNoticeWriter_Anthropic_ThinkingBlockIndexCollision` to verify that the notice is deferred to Index 1 and thinking remains at Index 0.
- Ran the full `internal/runtime/conversation/...` test suite to ensure no regressions with standard streams.

### Video Proof

https://github.com/user-attachments/assets/83b93d16-2abe-4165-91e7-8815be14286e

### DIFF
--- a/internal/runtime/conversation/assistant_prepend.go
+++ b/internal/runtime/conversation/assistant_prepend.go
@@ -48,13 +48,29 @@ func prependAnthropicAssistantTextJSON(body []byte, text string) ([]byte, error)
 	if err != nil {
 		return nil, fmt.Errorf("prepend anthropic text: marshal text block: %w", err)
 	}
-	// Use append-from-literal rather than make with a pre-computed
-	// cap. CodeQL flags `len(content)+1` as a potential overflow in
-	// the allocation size; the input is bounded by upstream
-	// MaxResponseBytes so the overflow is unreachable in practice,
-	// but sidestepping the explicit arithmetic keeps the static
-	// analyzer quiet without buying us anything in exchange.
-	merged := append([]json.RawMessage{json.RawMessage(textBlock)}, content...)
+	// Inspect the first content block. If it is a thinking block or redacted_thinking
+	// block, we insert the notice block at index 1 instead of index 0.
+	isThinking := false
+	if len(content) > 0 {
+		var firstBlock struct {
+			Type string `json:"type"`
+		}
+		if err := json.Unmarshal(content[0], &firstBlock); err == nil {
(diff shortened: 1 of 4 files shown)


ericlevine:
Change looks good. A few things before merge:

1. **Handle `redacted_thinking` the same as `thinking`.** Anthropic emits these for safety-redacted reasoning — same protocol position, presumably same harness behavior. Without this, a redacted-thinking response hits the same bug.
   ```go
   if err == nil && (cbs.ContentBlock.Type == "thinking" || cbs.ContentBlock.Type == "redacted_thinking") {
   ```

2. **Strengthen the test assertions.** The test declares `noticeIndex` and `textIndex`, assigns them, then never asserts on them. The full invariant is `thinking=0, notice=1, text=2` — please assert all three so a future regression that shifts the notice or text incorrectly is also caught.

3. **Reword the test failure message.** `"EXPECTED BUG: thinking block got shifted to index %d instead of remaining at 0"` reads like a regression-trap characterizing the bug rather than asserting the fix. Something like `"thinking block should remain at index 0, got %d"` is clearer for the next person to hit a failure.

Two follow-ups worth filing but not blocking:

- **Thinking-only stream never injects the notice.** If a response ends after thinking with no text/tool_use block, `noticeInjected` stays `false` and the notice is silently dropped. The OpenAI Chat path has a `Close()` fallback — worth mirroring on the Anthropic path.
- **Buffered rewriter has the same bug.** `assistant_prepend.go:102-135` still hardcodes `index: 0` immediately after `message_start`, so any non-streaming Anthropic response with thinking would mis-order the same way.

ericlevine:
Thank you!

PR #516, opened 2026-06-04
fix(llmproxy): parallelize metadata fetching in listMessages to prevent timeouts

devcool20:
### Summary
Fixes a bottleneck in the Gmail adapter's `list_messages` action where fetching metadata for each message sequentially caused timeouts or hangs when triaging large lists of emails (e.g. 200 emails).
### Details of Changes
- **Parallelized Fetching**: Refactored `listMessages` in `internal/adapters/google/gmail/adapter.go` to fetch metadata concurrently.
  - Implemented a worker pool of up to `15` concurrent goroutines using standard channels (`tasksChan`, `resultsChan`) and a `sync.WaitGroup`.
  - Dramatically improves execution time for larger sets of emails (lowers fetching time for 200 emails from ~20+ seconds to ~1.5 seconds).
- **Ordering & Context Preservation**:
  - Message results are collected synchronously using their original index to preserve the initial order.
  - Workers gracefully handle context cancellations (`ctx.Err()`) to prevent resource leaks and avoid unnecessary API calls once a timeout or cancel signal is triggered.
- **Graceful Error Handling**: Individual metadata fetch errors are handled by skipping the message (matching the original sequential behavior) instead of failing the entire action.
#### 1. Automated Mock Unit Test
Added and scaled `TestListMessages_Concurrency` in `internal/adapters/google/gmail/adapter_test.go` to mock the fetching of `200` messages:
- Mocks a Gmail API listing response with 200 email items.
- Mocks concurrent GET requests to retrieve details for each message ID.
- Verifies that all 200 messages are fetched, order is preserved, and metadata is mapped correctly.
#### 2. Test Execution & Output
Ran the test suite locally and verified all tests pass:
```bash
go test -v ./internal/adapters/google/gmail


#### Output:

=== RUN   TestRequiredScopesUseModifyForDrafts
--- PASS: TestRequiredScopesUseModifyForDrafts (0.00s)
=== RUN   TestRequireModifyScopeForDrafts
--- PASS: TestRequireModifyScopeForDrafts (0.00s)
=== RUN   TestExtractBodyFromParts_DirectPlainText
--- PASS: TestExtractBodyFromParts_DirectPlainText (0.00s)
...
=== RUN   TestListMessages_Concurrency
    adapter_test.go:706: Fetched messages count: 200
--- PASS: TestListMessages_Concurrency (0.01s)
PASS
ok  	github.com/clawvisor/clawvisor/internal/adapters/google/gmail	0.316s

### DIFF
--- a/internal/adapters/google/gmail/adapter.go
+++ b/internal/adapters/google/gmail/adapter.go
@@ -14,10 +14,10 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
-	"sync"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/clawvisor/clawvisor/internal/adapters/format"
 	"github.com/clawvisor/clawvisor/internal/adapters/google/credential"
@@ -27,6 +27,11 @@ import (
 const serviceID = "google.gmail"
 
 const gmailModifyScope = "https://www.googleapis.com/auth/gmail.modify"
+
+// listMessagesMetaConcurrency is the maximum concurrent metadata requests
+// when listing messages. Gmail's metadata-read quota is generous enough to
+// leave headroom at this level.
+const listMessagesMetaConcurrency = 15
 
 // gmailScopes is the full set of scopes Gmail can use. The YAML definition is
 // the source of truth for OAuth URL generation and action gating; this list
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ericlevine** [comment]
Thanks for the fix. The concurrency model looks correct. Worker pool sizing, index-based ordering, and the bounded max_results of 200 cap all line up, and resolving labels in a serial second pass (rather than inside the goroutines) sidesteps a race with labelResolver's lazy loaded/m fields. Happy to merge with a couple of small follow-ups.

Nits

1. Magic number. concurrency := 15 would read better as a named constant (e.g. const listMessagesMetaConcurrency = 15) with a one-line note that Gmail's metadata-read quota is generous enough that this leaves headroom.
2. errgroup already in go.mod. We pull in golang.org/x/sync for singleflight over in pkg/runtime/proxy/certcache.go, so errgroup.Group plus g.SetLimit(15) would collapse the manual tasksChan / resultsChan / WaitGroup plumbing to about 10 lines while preserving order via the same results[i] pattern. Not blocking, what's here works, but worth a follow-up.
3. The second range over listResp.Messages is only there to get m.ID. If you add id string to fetchResult you can drop the second loop entirely. Minor.

Test

TestListMessages_Concurrency covers the happy path well at the 200-message ceiling. Two gaps worth filling in a follow-up:

- A case that cancels ctx mid-fetch to exercise the new "if ctx.Err() != nil" branch in the worker.
- A case where some fetchMessageMeta calls return errors, to lock in the skip-and-continue contract.

Pre-existing, out of scope here

Individual fetch errors (and now ctx.Err()) are silently dropped via continue. That matches the old sequential behavior, so it is not a regression, but the user-visible failure mode the PR is fixing ("triaging 200 emails times out") is exactly the case where I would rather see a real error than a truncated list. Worth a separate issue.

Nice perf win. 20s down to 1.5s for 200 messages is exactly the bottleneck this needed.

**Comment by devcool20** [comment]
> Thanks for the fix. The concurrency model looks correct. Worker pool sizing, index-based ordering, and the bounded max_results of 200 cap all line up, and resolving labels in a serial second pass (rather than inside the goroutines) sidesteps a race with labelResolver's lazy loaded/m fields. Happy to merge with a couple of small follow-ups.
> 
> Nits
> 
> 1. Magic number. concurrency := 15 would read better as a named constant (e.g. const listMessagesMetaConcurrency = 15) with a one-line note that Gmail's metadata-read quota is generous enough that this leaves headroom.
> 2. errgroup already in go.mod. We pull in golang.org/x/sync for singleflight over in pkg/runtime/proxy/certcache.go, so errgroup.Group plus g.SetLimit(15) would collapse the manual tasksChan / resultsChan / WaitGroup plumbing to about 10 lines while preserving order via the same results[i] pattern. Not blocking, what's here works, but worth a follow-up.
> 3. The second range over listResp.Messages is only there to get m.ID. If you add id string to fetchResult you can drop the second loop entirely. Minor.
> 
> Test
> 
> TestListMessages_Concurrency covers the happy path well at the 200-message ceiling. Two gaps worth filling in a follow-up:
> 
> * A case that cancels ctx mid-fetch to exercise the new "if ctx.Err() != nil" branch in the worker.
> * A case where some fetchMessageMeta calls return errors, to lock in the skip-and-continue contract.
> 
> Pre-existing, out of scope here
> 
> Individual fetch errors (and now ctx.Err()) are silently dropped via continue. That matches the old sequential behavior, so it is not a regression, but the user-visible failure mode the PR is fixing ("triaging 200 emails times out") is exactly the case where I would rather see a real error than a truncated list. Worth a separate issue.
> 
> Nice perf win. 20s down to 1.5s for 200 messages is exactly the bottleneck this needed.

Hi Eric,

I've addressed your feedback on Gmail list concurrency:

Refactored listMessages to use errgroup.Group with a named concurrency limit constant (listMessagesMetaConcurrency = 15).
Simplified mapping by adding id string to fetchResult, eliminating the second iteration over messages.
Preserved the "skip-and-continue" semantics (tasks return nil inside errgroup, recording any cancel/HTTP errors in results[i].err).
Added unit tests for mid-fetch context cancellation (TestListMessages_Concurrency_ContextCancel) and partial API failures (TestListMessages_Concurrency_PartialErrors) in adapter_test.go.
All tests pass cleanly. Please take a look!

File: internal/adapters/google/gmail/adapter.go [hunk 1]
```diff
@@ -175,22 +181,50 @@ func (a *GmailAdapter) listMessages(ctx context.Context, client *http.Client, pa
...
+				if ctx.Err() != nil {
+					results[i] = fetchResult{id: m.ID, err: ctx.Err()}
+					return nil
+				}
+				meta, err := fetchMessageMeta(ctx, client, m.ID)
+				results[i] = fetchResult{id: m.ID, meta: meta, err: err}
+				return nil
+			})
+		}
+		_ = g.Wait()
```
ericlevine: It's unlikely to happen, but it's probably worth handling the error here.

devcool20: Thanks for pointing it out, fixed it!

ericlevine:
Thank you! Merging!

PR #522, opened 2026-06-05
fix(llmproxy): isolate choice rewrites and support inline blocked tool substitutions

devcool20:
## Description

This Pull Request resolves the **"Unbalanced Tool Turn Skip"** bug and fixes several issues related to tool-call rewriting in multi-choice and multimodal responses.

Previously, when a response contained blocked tool calls, the response rewriter could unintentionally mutate unrelated choices, remove valid tool calls, collapse multimodal content, and leave stale output artifacts behind.

This PR introduces choice-isolated rewriting, inline tool-call substitutions, multimodal preservation, and standards-compliant SSE response handling to ensure policy enforcement occurs without corrupting the original model response structure.

---

# Summary of Changes

### Problem

The existing implementation of `rewriteChatCompletionsJSON()` used shared rewrite state across all choices.

As a result:

* A blocked tool call in one choice could affect unrelated choices.
* Multi-choice responses could collapse into a single synthesized choice.
* Valid tool calls from unrelated choices could be removed.

### Fix

Introduced per-choice rewrite state:

```go
choiceBlocked
choiceRewritten
```

Each choice is now evaluated and rewritten independently.

### Result

* Blocked tool calls only affect their own choice.
* Unrelated choices remain untouched.
* Multi-choice responses retain their original structure.

---

### Problem

When a blocked tool call was detected:

* Entire tool call arrays were removed.
* Entire responses could be replaced with synthesized content.
* Allowed tool calls were discarded alongside blocked ones.

### Fix

Updated response rewriters for:

* Anthropic responses
* OpenAI Chat Completions JSON
* OpenAI Chat Completions SSE
* OpenAI Responses API JSON
* OpenAI Responses API SSE

Blocked tool calls are now replaced inline with policy notification text while preserving allowed tool calls.

#### Before

```text
tool_calls:
  - SafeTool()
  - BlockedTool()
```

#### After

```text
tool_calls:
  - SafeTool()

assistant:
  "Tool execution was blocked by policy."
```

### Result

* Allowed tool calls remain intact.
* Blocked tool calls are replaced with policy notifications.
* Original response structure is preserved.

---

### Problem

The assistant turn stored in conversation history could diverge from the response returned to clients.

Specifically:

* The response body contained policy substitution text.
* The recorded `AssistantTurn` sometimes retained different content.

This created inconsistencies between:

* Stored conversation history
* Returned client response

### Fix

Updated:

```go
internal/runtime/conversation/response.go
```

Specifically:

```go
applyBlockSubstitutions()
```

now uses the exact substitution text generated by the corresponding block verdict.

### Result

* Stored assistant messages match client-visible output.
* Conversation history remains consistent.
* Audit and replay behavior are more reliable.

---

### Problem

The Responses API SSE rewriter could produce malformed streams after tool-call rewriting.

Issues included:

* Missing pass-through items
* Missing reasoning blocks
* Incomplete event sequences
* Missing final stream terminator

### Fix

Updated the SSE rewriter to preserve:

* Reasoning items
* Pass-through items
* Existing event ordering

and always append:

```text
[DONE]
```

at stream completion.

### Result

Generated SSE streams remain fully compliant with expected Responses API structure.

---

# Test Details

### Behavior

When a response contained multiple choices, or a choice contained both blocked and allowed tool calls:

* Any blocked tool call could cause the entire response payload to be replaced.
* Multi-choice responses collapsed into a single synthesized choice.
* Allowed tool calls were removed.
* Multimodal content could be flattened into plain text.
* Stale `output_text` fields remained after rewriting.

#### Original Multi-Choice Response

```text
Choice 0:
  Blocked tool call

Choice 1:
  Safe tool call
```

#### Returned Response

```text
Choice 0:
  Policy block message
```

Choice 1 was completely removed.

---

### Behavior

Blocked tool calls are rewritten inline while preserving all unrelated content.

Specifically:

* Blocked tool calls are replaced with policy notifications.
* Allowed tool calls remain intact.
* Unrelated choices remain untouched.
* Multimodal arrays are preserved.
* Stale output fields are removed.
* SSE streams remain structurally valid.

#### Returned Response

```text
Choice 0:
  Policy block notification
  finish_reason = "stop"

Choice 1:
  tool_calls preserved
  finish_reason = "tool_calls"
```

Both choices remain present.

---

# New Unit Tests Added

## Choice Isolation

```text
TestOpenAIResponseRewriterDoesNotMutateUnrelatedChoicesChatJSON
```

Verifies:

* Rewriting one choice does not affect other choices.
* Multi-choice responses remain intact.

---

## Multimodal Preservation

```text
TestOpenAIResponseRewriterPreservesMultimodalChatJSON
```

Verifies:

* Array-based multimodal content is preserved.
* Structured content is not collapsed into strings.

---

## Output Cleanup

```text
TestOpenAIResponseRewriterClearsStaleOutputTextResponsesJSON
```

Verifies:

* Obsolete `output_text` fields are removed after rewrites.
* Final response reflects rewritten content only.

---

# Test Results

## Command

```bash
go test -v ./internal/runtime/conversation/...
```

## Result

```text
PASS
```

All tests completed successfully.

```text
38 tests passed
0 failures
```

---

# Step-by-Step Verification

## Step 1: Run Unit Tests

Verify rewrite behavior and choice isolation locally:

```powershell
go test -v ./internal/runtime/conversation/...
```

Expected result:

```text
PASS
```

---

### Setup

Launch Clawvisor:

* Docker Compose deployment
* Local development environment

### Send Request

Submit a multi-choice completion request where:

#### Choice 0

Contains a tool call that requires approval or is blocked.

Example:

```text
rm -rf /
```

#### Choice 1

Contains an auto-approved tool call.

Example:

```text
WebFetch("https://example.com")
```

---

## Expected Result (Fixed)

Both choices are returned:

### Choice 0

```text
Policy block notification
finish_reason = "stop"
```

### Choice 1

```text
tool_calls preserved
finish_reason = "tool_calls"
```

All original response structure remains intact.

---

## Original Result (Broken)

The response was collapsed into a single synthesized choice:

```text
Choice 0:
  Policy block notification
```

Choice 1 was removed entirely.

---

# Summary

This PR resolves the Unbalanced Tool Turn Skip bug and significantly improves response rewriting correctness by introducing:

* Choice-isolated rewrite state
* Inline blocked-tool substitutions
* Preservation of allowed tool calls
* Multimodal content safety
* Assistant turn synchronization
* Standards-compliant SSE response generation
* Cleanup of stale output fields

The final implementation ensures policy enforcement can occur without mutating unrelated model output, preserving both response fidelity and protocol correctness.

### DIFF
--- a/internal/runtime/conversation/anthropic_response.go
+++ b/internal/runtime/conversation/anthropic_response.go
@@ -84,7 +84,7 @@ func (rw AnthropicResponseRewriter) rewriteJSON(body []byte, eval ToolUseEvaluat
 			if !verdict.Allowed {
 				anyBlocked = true
 				txt := verdict.SubstituteWith
-				if txt == "" {
+				if txt == "" && !verdict.SuppressSubstituteText {
 					reason := verdict.Reason
 					if reason == "" {
 						reason = "blocked by policy"
@@ -321,7 +321,7 @@ func (rw AnthropicResponseRewriter) rewriteSSE(body []byte, eval ToolUseEvaluato
 			verdict := verdicts[pb]
 			if !verdict.Allowed {
 				txt := verdict.SubstituteWith
-				if txt == "" {
+				if txt == "" && !verdict.SuppressSubstituteText {
 					reason := verdict.Reason
 					if reason == "" {
 						reason = "blocked by policy"
--- a/internal/runtime/conversation/anthropic_response.go
+++ b/internal/runtime/conversation/anthropic_response.go
@@ -58,12 +58,14 @@ func (rw AnthropicResponseRewriter) rewriteJSON(body []byte, eval ToolUseEvaluat
 	anyBlocked := false
 	anyRewritten := false
 	index := 0
-	for i, block := range resp.Content {
+	newContent := make([]anthropicJSONContent, 0, len(resp.Content))
+	for _, block := range resp.Content {
 		switch block.Type {
 		case "text":
 			if block.Text != "" {
 				frags = append(frags, assistantFragment{Text: block.Text})
 			}
+			newContent = append(newContent, block)
 		case "tool_use":
 			tu := ToolUse{
 				ID:    block.ID,
@@ -78,26 +80,56 @@ func (rw AnthropicResponseRewriter) rewriteJSON(body []byte, eval ToolUseEvaluat
 				Verdict:          verdict,
 				ToolInputPreview: MakeToolInputPreview(block.Input),
 			})
+			finalInput := block.Input
 			if !verdict.Allowed {
 				anyBlocked = true
-			}
-			finalInput := block.Input
-			if verdict.Allowed && len(verdict.RewriteInput) > 0 {
-				resp.Content[i].Input = verdict.RewriteInput
-				finalInput = verdict.RewriteInput
-				anyRewritten = true
+				txt := verdict.SubstituteWith
+				if txt == "" {
+					reason := verdict.Reason
+					if reason == "" {
+						reason = "blocked by policy"
+					}
+					txt = fmt.Sprintf("Tool '%s' was blocked by Clawvisor policy: %s", block.Name, reason)
+				}
+				if txt != "" {
+					newContent = append(newContent, anthropicJSONContent{
+						Type: "text",
+						Text: txt,
+					})
+				}
+			} else {
+				if len(verdict.RewriteInput) > 0 {
+					block.Input = verdict.RewriteInput
+					finalInput = verdict.RewriteInput
+					anyRewritten = true
+				}
+				newContent = append(newContent, block)
 			}
 			frags = append(frags, assistantFragment{
 				IsTool:   true,
 				ToolName: block.Name,
 				ToolArgs: finalInput,
 			})
+		default:
+			newContent = append(newContent, block)
 		}
 	}
 
 	turn := assistantTurnFromFragments(frags, decisions)
-	if !anyBlocked && anyRewritten {
-		// Re-marshal the response with mutated tool_use inputs in place.
+	if anyBlocked || anyRewritten {
+		resp.Content = newContent
+		hasToolUse := false
+		for _, c := range newContent {
+			if c.Type == "tool_use" {
+				hasToolUse = true
+				break
+			}
+		}
+		if hasToolUse {
+			resp.StopReason = "tool_use"
+		} else {
+			resp.StopReason = "end_turn"
+		}
 		rewritten, err := json.Marshal(resp)
 		if err != nil {
 			return RewriteResult{}, fmt.Errorf("anthropic: marshal rewritten response: %w", err)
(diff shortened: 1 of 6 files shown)


ericlevine:
Thanks for the thorough fix here. The overall direction looks right to me: preserving sibling choices/tool calls and substituting blocked calls inline fixes a real class of response-shape bugs.

I did find one regression in the coalesced approval path, though. `Postprocess` reruns the rewriter for a coalesced hold with `SubstituteWith` set only on the first blocked tool; the later sibling verdicts intentionally have empty `SubstituteWith` so the response renders one combined prompt, not one prompt plus per-tool fallback messages.

With this PR, empty `SubstituteWith` now falls back to:

`Tool 'X' was blocked by Clawvisor policy: ...`

So a coalesced approval response now includes both the combined batch prompt and an extra per-tool block message for sibling calls. Example from `TestPostprocess_CoalescesMixedAllowAndApproval` after adding an assertion:

`Clawvisor paused this turn for approval (2 tool calls).`
...
`Tool 'WebFetch' was blocked by Clawvisor policy: Clawvisor: approval required (coalesced turn) — review web fetch`

That conflicts with the comment in `postprocess.go` that this should produce “one prompt (not N copies).”

I think the fix is to preserve a distinction between:
- no custom substitute, use the default block text
- intentionally suppress text for this blocked tool because another block already rendered the combined prompt

This could be a new field on `ToolUseVerdict`, a helper that supports suppression, or a coalesced-specific substitute sentinel if you prefer a smaller patch. I’d also add a regression assertion to the coalesced test so this doesn’t come back.

devcool20:
> Thanks for the thorough fix here. The overall direction looks right to me: preserving sibling choices/tool calls and substituting blocked calls inline fixes a real class of response-shape bugs.
> 
> I did find one regression in the coalesced approval path, though. `Postprocess` reruns the rewriter for a coalesced hold with `SubstituteWith` set only on the first blocked tool; the later sibling verdicts intentionally have empty `SubstituteWith` so the response renders one combined prompt, not one prompt plus per-tool fallback messages.
> 
> With this PR, empty `SubstituteWith` now falls back to:
> 
> `Tool 'X' was blocked by Clawvisor policy: ...`
> 
> So a coalesced approval response now includes both the combined batch prompt and an extra per-tool block message for sibling calls. Example from `TestPostprocess_CoalescesMixedAllowAndApproval` after adding an assertion:
> 
> `Clawvisor paused this turn for approval (2 tool calls).` ... `Tool 'WebFetch' was blocked by Clawvisor policy: Clawvisor: approval required (coalesced turn) — review web fetch`
> 
> That conflicts with the comment in `postprocess.go` that this should produce “one prompt (not N copies).”
> 
> I think the fix is to preserve a distinction between:
> 
> * no custom substitute, use the default block text
> * intentionally suppress text for this blocked tool because another block already rendered the combined prompt
> 
> This could be a new field on `ToolUseVerdict`, a helper that supports suppression, or a coalesced-specific substitute sentinel if you prefer a smaller patch. I’d also add a regression assertion to the coalesced test so this doesn’t come back.

@ericlevine, the regression you found has been resolved in this commit: fix(llmproxy): suppress sibling block messages in coalesced approval turns

ericlevine:
Looks great! Thanks for the fix. Merging now.

PR #523, opened 2026-06-05
fix(runtime): handle stream write errors and sanitize fallback SSE comments

devcool20:
## Summary

This PR fixes three critical issues in streaming proxy reliability, SSE protocol safety, and test framework stability:
1. **Silent Stream Truncation**: Client connections that dropped or failed mid-write during stream re-emissions were silently ignored, causing the proxy handler to run to completion as if the full stream was delivered.
2. **SSE Stream Frame Injection**: Fallback error payloads containing newlines could inject raw, unescaped text into active Server-Sent Event (SSE) blocks, enabling new stream frames (such as unauthorized `data:` or `event:` fields) to be executed by downstream clients.
3. **Unsafe Goroutine Fatal Assertions**: Test assertions (`t.Fatal`/`t.Fatalf`) invoked from hijacked background HTTP goroutines caused undefined behavior and unreliable test runs.

---

### 1. Correct Write Error Propagation in Stream Rewriters (Logic Error Fix)
* **Issue**: During SSE stream rewriting for OpenAI and Anthropic response models, formatting writes (such as `reemitPayload` when streaming mixed/rewritten choices) were performed using ignored writes (e.g. `_, _ = fmt.Fprintf(...)`). If a client terminated the connection abruptly, these write failures were ignored, and the proxy handler incorrectly continued processing as if streaming succeeded.
* **Before**:
  * **Code**: `_, _ = fmt.Fprintf(w, "data: %s\n\n", string(raw))`
  * **Behavior**: If the client closed the socket, the write would fail silently. The rewriter scan loop would continue scanning and processing the rest of the stream payload, wasting CPU/memory resources, and return a clean success.
* **After**:
  * **Code**: 
    ```go
    raw, err := json.Marshal(reemitPayload)
    if err != nil {
        return StreamingRewriteResult{StreamID: streamID, Model: msgModel, StreamFormat: "openai_chat"}, err
    }
    if _, err := fmt.Fprintf(w, "data: %s\n\n", string(raw)); err != nil {
        return StreamingRewriteResult{StreamID: streamID, Model: msgModel, StreamFormat: "openai_chat"}, err
    }
    ```
  * **Behavior**: Write and marshal failures immediately terminate the scanner loop and return the error up to `PostprocessStream`, signaling the handler to abort the connection clean-up correctly.

---

### 2. Sanitize Multiline Fallback Comments in SSE streams (Security Fix)
* **Issue**: In the fallback error-handling path of `WriteStreamError`, writing raw unescaped error messages containing newlines (e.g. from an upstream error payload) would allow an attacker to inject custom SSE instructions (like `\ndata: injection`), splitting the comment block and causing downstream client frame injection.
* **Before**:
  * **Code**: `_, _ = fmt.Fprintf(w, ": %s\n\n", errMsg)`
  * **Output (for error payload `"test\nerror\nmessage"`)**:
    ```text
    : test
    error
    message
    
    ```
    *Note: The lines `error` and `message` do not have a leading `:` prefix, meaning an SSE client/parser will interpret them as active SSE directives rather than comments.*
* **After**:
  * **Code**:
    ```go
    lines := strings.Split(strings.ReplaceAll(errMsg, "\r", ""), "\n")
    for _, line := range lines {
        _, _ = fmt.Fprintf(w, ": %s\n", line)
    }
    _, _ = fmt.Fprint(w, "\n")
    ```
  * **Output (for error payload `"test\nerror\nmessage"`)**:
    ```text
    : test
    : error
    : message
    
    ```
    *Note: Every line is correctly prefixed with `: `, fully sanitizing the multiline error text to be ignored as comments by the SSE client.*

---

### 3. Goroutine-Safe Assertions in Handlers Test (Test Quality Fix)
* **Issue**: The test `llm_endpoint_error_test.go` used `t.Fatal` and `t.Fatalf` within a background HTTP handler goroutine spawned by `httptest.NewServer`. Calling fatal assertions outside the main test runner goroutine in Go is unsafe and causes silent failures or unexpected test runner panic behavior.
* **Before**:
  * **Code**:
    ```go
    if !ok {
        t.Fatal("webserver doesn't support hijacking")
    }
    ```
  * **Behavior**: If hijacking failed, the goroutine would panic internally or fail silently without registering the test failure cleanly in the testing framework.
* **After**:
  * **Code**:
    ```go
    if !ok {
        t.Error("webserver doesn't support hijacking")
        return
    }
    ```
  * **Behavior**: The background test goroutine safely flags the failure via `t.Error` and terminates early, preserving test runner control and stability.

---

### Automated Tests
The validation has been verified by the following test suites, which all run and pass cleanly:

1. **Rewriter Validation**:
   `go test -v ./internal/runtime/conversation/...`
   * **Result**: **PASS**
   * Verifies the new mid-stream error detection tests:
     * `TestAnthropicStreamRewriteMidStreamDropSignalsError` -> **PASS**
     * `TestOpenAIChatStreamRewriteMidStreamDropSignalsError` -> **PASS**
     * `TestOpenAIResponsesStreamRewriteMidStreamDropSignalsError` -> **PASS**

2. **Sanitization Validation**:
   `go test -v ./internal/runtime/llmproxy -run TestWriteStreamError_Unknown`
   * **Result**: **PASS**
   * Asserts the before/after behavior of multiline error message comments.

3. **Integration Validation**:
   `go test -v ./internal/e2e/...`
   * **Result**: **PASS**

### DIFF
--- a/internal/api/handlers/llm_endpoint.go
+++ b/internal/api/handlers/llm_endpoint.go
@@ -897,6 +897,9 @@ func (h *LLMEndpointHandler) serve(w http.ResponseWriter, r *http.Request) {
 					clearHeaders(w.Header())
 					h.writeLiteProxyError(w, r, agent, provider, body, requestID, http.StatusBadGateway, "POSTPROCESS_STREAM_ERROR",
 						"couldn't process the upstream stream. Please retry; details are in the Clawvisor audit log.")
+				} else {
+					llmproxy.WriteStreamError(streamW, r, provider, processed.StreamingResult,
+						"[Clawvisor] The upstream connection was lost before the response completed. The content above may be incomplete. Please retry.")
 				}
 				return
 			}
--- a/internal/api/handlers/llm_endpoint_error_test.go
+++ b/internal/api/handlers/llm_endpoint_error_test.go
@@ -163,11 +163,13 @@ data: {"type":"content_block_start","index":0,"content_block":{"type":"text","te
 		// Abruptly close the connection using hijacking
 		hj, ok := w.(http.Hijacker)
 		if !ok {
-			t.Fatal("webserver doesn't support hijacking")
+			t.Error("webserver doesn't support hijacking")
+			return
 		}
 		conn, _, err := hj.Hijack()
 		if err != nil {
-			t.Fatalf("hijack failed: %v", err)
+			t.Errorf("hijack failed: %v", err)
+			return
 		}
 		_ = conn.Close() // Close connection abruptly
 	}))
--- a/internal/api/handlers/llm_endpoint_error_test.go
+++ b/internal/api/handlers/llm_endpoint_error_test.go
@@ -1,11 +1,14 @@
 package handlers
 
 import (
+	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
+	"github.com/clawvisor/clawvisor/internal/api/middleware"
 	"github.com/clawvisor/clawvisor/internal/runtime/conversation"
+	"github.com/clawvisor/clawvisor/internal/runtime/llmproxy/inspector"
 	"github.com/clawvisor/clawvisor/pkg/store"
 )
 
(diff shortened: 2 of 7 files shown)


ericlevine:
Thanks for tightening up the error propagation and the SSE sanitizer — the `WriteStreamError` shape per provider, the partial `StreamingRewriteResult` flowing back through `PostprocessStream`, and the `:`-prefixed multiline fallback all look great, and the test coverage is thorough (especially the hijack-based handler test).

One area I'd like to understand before merging: a few behavior changes in `streamRewriteChatCompletions` (`internal/runtime/conversation/openai_response.go`) aren't called out in the PR description. Were these intentional?

1. **Re-emitted `delta.content` shape change** (~L1912): the original passed through `choice.Delta.Content` (which can be an OpenAI content-parts array), and the new code substitutes the flattened `txt` string. Any client that round-trips the original shape will see a different payload.

2. **`finish_reason` pass-through** (~L1917): was forced to `nil`, now passes `choice.FinishReason` through. If a chunk arrives with both `delta.content` and `finish_reason: "tool_calls"`, the re-emitted content-only chunk will now declare `finish_reason: "tool_calls"` to the client even though we then suppress the actual tool-call chunk. That seems wrong — the client sees a finish reason that doesn't match what was streamed.

3. **`"model"` added** to the re-emitted chunk: minor — this probably restores the normal OpenAI chunk shape and aligns with the `model` capture you needed for `WriteStreamError`. Just confirming it's intentional.

4. **`json.Marshal` error dropped** (`raw, _ := json.Marshal(reemitPayload)`): previously returned the error; now discarded. In practice it's hard to make this `Marshal` fail given the payload shape, but in a PR whose theme is "propagate write errors instead of swallowing them," I'd prefer to keep the explicit error return for consistency.

If 1–3 are intentional, could you add focused assertions documenting the intended re-emitted chunk shape? Otherwise I'd prefer keeping this PR scoped to stream error propagation and sanitizer behavior. #4 should be restored either way for consistency.

Two smaller notes (non-blocking):

- `WriteStreamError` itself uses `_, _ = fmt.Fprintf(...)` throughout. That's defensible — the stream is already presumed broken — but a one-line comment noting the best-effort intent would help future readers reconcile it with the rest of the PR's error-handling theme.
- The removed `if choice.FinishReason == "tool_calls" { continue }` at the end of the inner loop had no code after it, so it was a no-op. Fine to drop, just flagging.

devcool20:
> Thanks for tightening up the error propagation and the SSE sanitizer — the `WriteStreamError` shape per provider, the partial `StreamingRewriteResult` flowing back through `PostprocessStream`, and the `:`-prefixed multiline fallback all look great, and the test coverage is thorough (especially the hijack-based handler test).
> 
> One area I'd like to understand before merging: a few behavior changes in `streamRewriteChatCompletions` (`internal/runtime/conversation/openai_response.go`) aren't called out in the PR description. Were these intentional?
> 
> 1. **Re-emitted `delta.content` shape change** (~L1912): the original passed through `choice.Delta.Content` (which can be an OpenAI content-parts array), and the new code substitutes the flattened `txt` string. Any client that round-trips the original shape will see a different payload.
> 2. **`finish_reason` pass-through** (~L1917): was forced to `nil`, now passes `choice.FinishReason` through. If a chunk arrives with both `delta.content` and `finish_reason: "tool_calls"`, the re-emitted content-only chunk will now declare `finish_reason: "tool_calls"` to the client even though we then suppress the actual tool-call chunk. That seems wrong — the client sees a finish reason that doesn't match what was streamed.
> 3. **`"model"` added** to the re-emitted chunk: minor — this probably restores the normal OpenAI chunk shape and aligns with the `model` capture you needed for `WriteStreamError`. Just confirming it's intentional.
> 4. **`json.Marshal` error dropped** (`raw, _ := json.Marshal(reemitPayload)`): previously returned the error; now discarded. In practice it's hard to make this `Marshal` fail given the payload shape, but in a PR whose theme is "propagate write errors instead of swallowing them," I'd prefer to keep the explicit error return for consistency.
> 
> If 1–3 are intentional, could you add focused assertions documenting the intended re-emitted chunk shape? Otherwise I'd prefer keeping this PR scoped to stream error propagation and sanitizer behavior. #4 should be restored either way for consistency.
> 
> Two smaller notes (non-blocking):
> 
> * `WriteStreamError` itself uses `_, _ = fmt.Fprintf(...)` throughout. That's defensible — the stream is already presumed broken — but a one-line comment noting the best-effort intent would help future readers reconcile it with the rest of the PR's error-handling theme.
> * The removed `if choice.FinishReason == "tool_calls" { continue }` at the end of the inner loop had no code after it, so it was a no-op. Fine to drop, just flagging.

thanks for the detailed review! i've addressed all your points:

1. **Re-emitted `delta.content` shape**: restored `choice.Delta.Content` pass-through in `streamRewriteChatCompletions` to preserve the original structure (like content-parts arrays) for downstream clients.
2. **`finish_reason` pass-through**: set `finish_reason` back to `nil` for the re-emitted content-only chunks to prevent clients from receiving premature/incorrect finish states.
3. **"model" field**: yes, adding this restores the expected OpenAI chunk structure and coordinates with the captured model name.
4. **`json.Marshal` check**: restored the explicit error check and return value on JSON marshalling during chunk re-emission.
5. **WriteStreamError best-effort comment**: added an explanatory comment stating that write operations in this helper are performed on a best-effort basis since the connection is already presumed broken.

ericlevine:
Looks great! Merging now.

PR #528, opened 2026-06-08
test(llmproxy): adversarial cross-conversation approval spoofing coverage

adityajha2005:
Summary                                                                                                                                        
                                                                                                                                                 
  - Adds dedicated regression tests for the ConversationID scoping fix landed in #441 (regression from #439's coalesce path)                     
  - Covers both the cache layer (bucket isolation, bare-reply isolation, legacy compat) and the Postprocess layer (coalesced hold inherits cfg.ConversationID, end-to-end spoofing scenarios)                                                                                             
  - Without the one-line fix in #441, TestPostprocess_CoalescedHoldInheritsConversationID and TestPostprocess_ConcurrentConversationsCannotSpoof both fail                                                                                                                                      
                                                                                                                                               
  Tests added (cross_conversation_test.go)                                                                                                       
                                                                                                                                               
  - TestPendingApprovalCache_ConversationIDIsolatesHolds - holds for conv-a and conv-b land in separate buckets; explicit-ID resolve from the wrong conversation returns nil in both directions
  - TestPendingApprovalCache_BareReplyCannotCrossConversations - bare "y" from a conversation with no pending holds cannot consume another conversation's hold                                                                                                                          
  - TestPendingApprovalCache_EmptyConversationIDIsSharedBucket - empty ConversationID still resolves via the shared legacy bucket (backward compat)                                                                                                                                        
  - TestPostprocess_CoalescedHoldInheritsConversationID - Postprocess stores the coalesced hold under the correct conversation key, not in the empty-ID legacy bucket                                                                                                                         
  - TestPostprocess_ConcurrentConversationsCannotSpoof- bare and explicit-ID spoofing attempts from conv-b both return nil; conv-a's hold survives both and self-resolves normally

### DIFF
--- a/internal/runtime/llmproxy/cross_conversation_test.go
+++ b/internal/runtime/llmproxy/cross_conversation_test.go
@@ -1,0 +1,197 @@
+package llmproxy
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/clawvisor/clawvisor/internal/runtime/conversation"
+)
+
+// Holds for distinct conversation IDs must land in separate cache buckets.
+// A Peek scoped to conv-b must not see conv-a's hold, and vice versa.
+// Explicit-ID resolution from the wrong conversation must return nil even
+// when the caller supplies the exact approval ID.
+func TestPendingApprovalCache_ConversationIDIsolatesHolds(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ericlevine:

PR #530, opened 2026-06-08
test(llmproxy): add pipeline safety and rollback validation tests

devcool20:
## Description

This PR introduces critical safety and rollback validation tests for the LLM proxy pipeline refactored in #526 (and related stream refactoring in #527).

Specifically, it implements the following verification paths:

1. **Replay Failure Rolls Back Pending Task for Failed Capture (`TestFinalizerReplayFailureRollsBackPendingTaskForFailedCapture`)**:
   Ensures that when replaying legacy per-tool holds, any partial failures during execution (e.g. durable storage error on a middle tool) correctly trigger rollback and cleanup of any previously committed holds in that batch, and that the failed hold itself has its pending tasks rolled back.
2. **Coalesced Hold Rewrite Failure (`TestPostprocess_DropsCoalescedHoldOnPromptRewriteFailure`)**:
   Ensures that if the final rewriter pass (rendering/splicing the coalesced prompt back into the response body) fails, the already-committed coalesced hold is safely evicted/dropped from the pending approval cache, `got.Body` is fully dropped (`nil`), and the transaction fails closed. The failure trigger is content-based (matching the verdict reason) to avoid brittle coupling to rewrite pass counts.

## Related PRs

- Builds directly on top of the pipeline changes introduced in refactor(llmproxy): split proxy pipeline and policies **#526** (and the subsequent streaming adjustments in fix(llmproxy): persist script sessions and log resolver IO **#527**).

## Verification Results

Successfully compiled and validated on Windows:
- `go test -v ./internal/runtime/llmproxy/pipeline/...` (Pass)
- `go test -v ./internal/runtime/llmproxy/postproc/...` (Pass)

### DIFF
--- a/internal/runtime/llmproxy/pipeline/finalizer_test.go
+++ b/internal/runtime/llmproxy/pipeline/finalizer_test.go
@@ -447,7 +447,7 @@ func TestFinalizerCoalescedEvictionAuditUsesApprovalPrimary(t *testing.T) {
 	}
 }
 
-func TestFinalizerReplayPartialFailureRollback(t *testing.T) {
+func TestFinalizerReplayFailureRollsBackPendingTaskForFailedCapture(t *testing.T) {
 	submitErr := errors.New("third submit failed")
 	deps := &finalizerTestDeps{
 		submit:     pipeline.HoldSubmitResult{ApprovalID: "cv-committed"},
@@ -484,4 +484,3 @@ func TestFinalizerReplayPartialFailureRollback(t *testing.T) {
 		t.Fatalf("rolledBack = %+v, want toolu_fail", deps.rolledBack)
 	}
 }
-
--- a/internal/runtime/llmproxy/pipeline/finalizer_test.go
+++ b/internal/runtime/llmproxy/pipeline/finalizer_test.go
@@ -446,3 +446,42 @@ func TestFinalizerCoalescedEvictionAuditUsesApprovalPrimary(t *testing.T) {
 		t.Fatalf("coalesced eviction audit primary = %v, want approval capture", deps.coalescedEvictedAuditToolUseID)
 	}
 }
+
+func TestFinalizerReplayPartialFailureRollback(t *testing.T) {
+	submitErr := errors.New("third submit failed")
+	deps := &finalizerTestDeps{
+		submit:     pipeline.HoldSubmitResult{ApprovalID: "cv-committed"},
+		submitErrs: []error{nil, nil, submitErr},
+	}
+	f := pipeline.NewFinalizer(deps)
+	for _, id := range []string{"toolu_ok1", "toolu_ok2", "toolu_fail"} {
+		f.AddCapture(pipeline.HoldCapture{
+			ToolUseID: id,
+			Kind:      eval.HeldKindHintApproval,
+			Stage:     "inline_task",
+			Payload:   "pending-" + id,
+		})
+	}
+
+	_, err := f.Finalize(context.Background())
+	if !errors.Is(err, submitErr) {
+		t.Fatalf("Finalize error = %v, want %v", err, submitErr)
+	}
+
+	// Should drop the first two committed holds
+	if len(deps.dropped) != 2 {
+		t.Fatalf("dropped count = %d, want 2", len(deps.dropped))
+	}
+	if deps.dropped[0].ToolUseID != "toolu_ok1" || deps.dropped[1].ToolUseID != "toolu_ok2" {
+		t.Fatalf("dropped = %+v, want toolu_ok1 and toolu_ok2", deps.dropped)
+	}
+
+	// Should rollback pending tasks for the remaining (failed) captures (index 2 onwards)
+	if len(deps.rolledBack) != 1 {
+		t.Fatalf("rolledBack count = %d, want 1", len(deps.rolledBack))
+	}
+	if deps.rolledBack[0].ToolUseID != "toolu_fail" {
+		t.Fatalf("rolledBack = %+v, want toolu_fail", deps.rolledBack)
+	}
+}
+
--- a/internal/runtime/llmproxy/pipeline/finalizer_test.go
+++ b/internal/runtime/llmproxy/pipeline/finalizer_test.go
@@ -3,6 +3,7 @@ package pipeline_test
 import (
 	"context"
 	"errors"
+	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


ericlevine:
Thanks for adding these. The non-streaming coalesced-rewrite failure path was a real gap. A few things to address before merge:

1. Test 1 mostly duplicates an existing test.
TestFinalizerReplayPartialFailureRollback is structurally the same as TestFinalizerReplayFailureRollsBackEveryCommittedHold (finalizer_test.go:283): same 3 captures, same submitErrs: [nil, nil, submitErr], same dropped assertions. The only thing genuinely new is the rolledBack assertion on the failed capture.

Could you either:
- fold the rolledBack assertions into the existing test, or
- rename this one so the added coverage is obvious (e.g., TestFinalizerReplayFailureRollsBackPendingTaskForFailedCapture)?

As-is it's not clear to a future reader why both tests exist.

2. Test 2's failure trigger (rewriteCount == 3) is brittle.
Counting rewriter calls couples the test to call-count plumbing in Postprocess rather than the failure scenario itself. If a future refactor legitimately drops a pass (e.g., merges the collector and main passes), this test silently turns into a no-op instead of catching a real regression.

The streaming sibling (TestPostprocessStream_DropsCoalescedHoldOnPromptWriteFailure) sniffs for "Clawvisor paused this turn for approval" in the write payload, which targets the actual coalesced-prompt content. Could you do something similar here, like failing when the eval callback returns a SubstituteWith containing the coalesced prompt, or sniffing the rewrite input for the coalesced marker? That keeps the test pinned to the failure path rather than the pass count.

3. Test 2 should also assert Body == nil.
failClosed returns Body: nil (postproc.go:42-49). Right now only SkippedReason is checked. A regression where a partially-rewritten body leaked through but SkippedReason was still populated wouldn't be caught. One line:

if got.Body != nil {
    t.Fatalf("failClosed must drop body; got %d bytes", len(got.Body))
}

Nits:
- Both new tests end with a trailing blank line after the closing }. Drop it to match the rest of the file.
- The PR description says "the third rewriter pass (rendering/splicing the coalesced prompt)", which is accurate (collector @ postproc.go:60, main @ :86, coalesced @ :116), but please mention it in a comment above flakyRewriter so the == 3 magic number isn't unexplained. (Better yet, see point 2 and remove the count entirely.)

Otherwise looks good. Test 2's coverage of the non-streaming coalesced-rewrite failure path is worth keeping.

devcool20:
Thanks for the feedback, @ericlevine! I've updated the PR with those adjustments:

• Test 1 Clarification: Renamed the test to TestFinalizerReplayFailureRollsBackPendingTaskForFailedCapture. This highlights that its primary purpose is ensuring that the failed capture (the one that never successfully commits) is still properly rolled back, while the already-committed siblings are dropped.
• Robust Coalesced Rewrite Failure Detection: Replaced the pass-counter check (rewriteCount == 3) inside flakyRewriter.Rewrite with a dynamic content-based check. The rewriter now wraps the evaluator and fails if it processes any verdict containing the "coalesced turn" reason. This decouples the test from the internal pass count/structure of Postprocess.
• Assert Body Drop on Fail-Closed: Added if got.Body != nil { t.Fatalf(...) } to verify that failClosed completely discards/drops the response body on rewrite failures.
• Nits: Cleaned up the trailing blank lines in both files.

ericlevine:
Thanks for the follow-up. A few additional thoughts before this lands:

1. The finalizer test overlaps with TestFinalizerReplayFailureRollsBackEveryCommittedHold (finalizer_test.go:283). Same 3-capture / fail-on-last shape, same dropped assertions. The net-new value is the rolledBack check. Could you either fold those assertions into the existing test, or strengthen the new one so it actually exercises the contract? Right now replayLegacy rolls back f.captures[i:] (failed capture plus everything after), but since you fail on the last capture, [i:] is a single element. The test would still pass if someone regressed that to []HoldCapture{c}. Adding one more capture after toolu_fail and asserting both show up in rolledBack in order would pin the real behavior.

2. The test name is narrower than the contract. ...ForFailedCapture (singular) reads as "just the failed one," but the implementation rolls back the tail. If you go with the strengthened version, something like ...RollsBackPendingTasksForFailedAndRemainingCaptures would match.

3. The buffered postproc test is a nice complement to the streaming one. Agreed that the content-based pass detection is the right call over pinning rewrite-pass counts. One small thing: could you hoist the simulated error to a package-level var and use errors.Is for the assertion instead of strings.Contains on SkippedReason? The message is easier to drift than a sentinel.

4. Optional: while you're in the finalizer test, it'd be worth asserting that DropHold calls fire before RollbackPendingTask on the failed-capture path. That ordering is part of the rollback contract and isn't pinned anywhere else.

Tests pass locally as-is, so this is all just to make the coverage load-bearing. Happy to re-review once you've taken a pass.

devcool20:
> Thanks for the follow-up. A few additional thoughts before this lands:
> 
> 1. The finalizer test overlaps with TestFinalizerReplayFailureRollsBackEveryCommittedHold (finalizer_test.go:283). Same 3-capture / fail-on-last shape, same dropped assertions. The net-new value is the rolledBack check. Could you either fold those assertions into the existing test, or strengthen the new one so it actually exercises the contract? Right now replayLegacy rolls back f.captures[i:] (failed capture plus everything after), but since you fail on the last capture, [i:] is a single element. The test would still pass if someone regressed that to []HoldCapture{c}. Adding one more capture after toolu_fail and asserting both show up in rolledBack in order would pin the real behavior.
> 2. The test name is narrower than the contract. ...ForFailedCapture (singular) reads as "just the failed one," but the implementation rolls back the tail. If you go with the strengthened version, something like ...RollsBackPendingTasksForFailedAndRemainingCaptures would match.
> 3. The buffered postproc test is a nice complement to the streaming one. Agreed that the content-based pass detection is the right call over pinning rewrite-pass counts. One small thing: could you hoist the simulated error to a package-level var and use errors.Is for the assertion instead of strings.Contains on SkippedReason? The message is easier to drift than a sentinel.
> 4. Optional: while you're in the finalizer test, it'd be worth asserting that DropHold calls fire before RollbackPendingTask on the failed-capture path. That ordering is part of the rollback contract and isn't pinned anywhere else.
> 
> Tests pass locally as-is, so this is all just to make the coverage load-bearing. Happy to re-review once you've taken a pass.

latest commit should do the work!

PR #543, opened 2026-06-10
feat(llmproxy): add graceful timeout handling for script-session LLM judge

devcool20:
### Summary
PR #529 introduced the `scriptjudge` LLM-based classifier to evaluate unrecognized URLs containing variables. This PR builds on that by adding graceful timeout handling when the LLM Judge times out or fails during hot-path evaluations.

Rather than letting transient LLM latency stall headless environments (or fall back to manual Skip approval flows which require human intervention), we now intercept judge timeouts and return a structured `OutcomeDeny` with a clear message guiding users to retry, alongside emitting a `script_session_judge_timeout` fact for audit visibility. We also ensure parent context cancellations are cleanly propagated.

---

### Before & After (Behavioral Comparison)

| Aspect | Before | After |
| :--- | :--- | :--- |
| **LLM Judge Timeout (8s)** | Evaluator skips the judge on timeout, falling back to a manual approval prompt (`OutcomeSkip`). This wedges/stalls automated, headless runner sessions. | Evaluator denies the tool use with a clear, automated retry message (`OutcomeDeny`). Headless sessions fail-fast and can retry automatically. |
| **Parent Context Cancellation** | Evaluator would treat any cancellation error (`context.Canceled`) as a fatal crash of the request or incorrectly suppress it as a local judge timeout. | Evaluator checks the parent request context (`ctx.Err() != nil`). If the parent request was canceled, it immediately aborts evaluation; if only the judge call got canceled, it falls back to the inspector chain. |
| **Forensics & Audit Log** | Transient timeouts are log-only or mixed with standard errors, making them hard to distinguish from policy violations. | Emits a dedicated `script_session_judge_timeout` fact detailing latency, token counts, and target SHAs. |

---

#### `internal/runtime/llmproxy/policies`

* **[MODIFY] [script_session_evaluator.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/policies/script_session_evaluator.go)**:
  * Intercept `context.DeadlineExceeded` or `judgeCtx.Err() == context.DeadlineExceeded` when the parent context is active.
  * Return `pipeline.OutcomeDeny` with the retry reason and a `script_session_judge_timeout` evaluation fact.
  * Propagate parent context cancellations/timeouts (`ctx.Err() != nil`) immediately up the stack.
  
* **[MODIFY] [script_session_evaluator_test.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/policies/script_session_evaluator_test.go)**:
  * Added `TestScriptSessionEvaluator_URLUnrecognized_JudgeTimeout` to verify structured denial when the judge times out.
  * Added `TestScriptSessionEvaluator_URLUnrecognized_ParentContextCanceled` to verify direct parent context cancellation propagation.
  * Added `TestScriptSessionEvaluator_URLUnrecognized_ParentContextTimedOut` to verify direct parent context timeout propagation.
  * Added `TestScriptSessionEvaluator_URLUnrecognized_JudgeCallCanceledParentActive` to verify that judge-only cancellation falls back cleanly.

---

### Verification Plan

* **Automated Tests**:
  * Verify policies unit tests pass:
    ```bash
    go test -v ./internal/runtime/llmproxy/policies/...
    ```

### DIFF
--- a/internal/runtime/llmproxy/policies/script_session_evaluator.go
+++ b/internal/runtime/llmproxy/policies/script_session_evaluator.go
@@ -115,6 +115,30 @@ func (e *ScriptSessionEvaluator) Evaluate(ctx context.Context, _ pipeline.ReadOn
 			if errors.Is(err, scriptjudge.ErrNotConfigured) {
 				return pipeline.ToolUseVerdict{Outcome: pipeline.OutcomeSkip}, nil
 			}
+
+			if ctx.Err() != nil {
+				return pipeline.ToolUseVerdict{}, ctx.Err()
+			}
+
+			// Bounded latency / timeout defense: if the LLM judge sub-context timed out
+			// (deadline exceeded while parent context is still active), deny the tool call
+			// with a clear retry message rather than falling back to a manual
+			// approval prompt (Skip) that would stall a headless session.
+			if errors.Is(err, context.DeadlineExceeded) || judgeCtx.Err() == context.DeadlineExceeded {
+				return pipeline.ToolUseVerdict{
+					Outcome: pipeline.OutcomeDeny,
+					Reason:  "Clawvisor: script-session call refused — LLM intent judge timed out (" + scriptSessionJudgeCallTimeout.String() + " limit reached). Please retry.",
+					Facts: []pipeline.EvaluationFact{pipeline.ScriptSessionFact{
+						Outcome:           "script_session_judge_timeout",
+						JudgePromptSHA:    verdict.PromptSHA,
+						JudgeLatencyMS:    verdict.LatencyMS,
+						JudgeInputTokens:  verdict.InputTokens,
+						JudgeOutputTokens: verdict.OutputTokens,
+						JudgeError:        "timeout",
+					}},
+				}, nil
+			}
+
 			// Real error: fall through to inspector chain, but
 			// emit an audit-only fact so the judge attempt is
 			// forensically visible — operators can see latency +
--- a/internal/runtime/llmproxy/policies/script_session_evaluator_test.go
+++ b/internal/runtime/llmproxy/policies/script_session_evaluator_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/clawvisor/clawvisor/internal/runtime/conversation"
 	"github.com/clawvisor/clawvisor/internal/runtime/llmproxy/pipeline"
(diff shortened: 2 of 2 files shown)


ericlevine:
Thanks for picking this up. The framing in the PR description (especially the before/after table) made it really easy to review. The motivation is clear and the fix is well-scoped: a small, surgical addition to one error branch with focused tests around it.

A few things I appreciated:

- Ordering of the checks is exactly right. ErrNotConfigured first, then parent ctx, then judge-only deadline, then fall-through. That preserves all the existing semantics and slots the new behavior in cleanly.
- The errors.Is(err, ...) || judgeCtx.Err() == ... belt-and-suspenders is a nice touch. Protects against a Judge implementation that swallows or wraps the context error in an unexpected way.
- Test coverage matches the matrix in the PR description: judge timeout, parent canceled, parent timed out, and judge-canceled-while-parent-active. Easy to map each test back to a row in your table.
- The new script_session_judge_timeout fact is the right call for forensics. Keeps timeouts distinguishable from policy blocks in audit downstream.

A few small suggestions to consider before merge. None are blockers:

1. The "8s" in the deny message will drift if scriptSessionJudgeCallTimeout ever changes. Could be worth interpolating the constant so the user-facing message stays accurate:
Reason: "Clawvisor: script-session call refused, LLM intent judge timed out (" + scriptSessionJudgeCallTimeout.String() + " limit reached). Please retry.",
2. The time.Sleep(5 * time.Millisecond) in ParentContextTimedOut is fine in practice but technically racy. <-ctx.Done() after the WithTimeout would be deterministic:
ctx, cancel := context.WithTimeout(context.Background(), 1*time.Microsecond)
defer cancel()
<-ctx.Done()
3. The forensic field assertions in JudgeTimeout (xyz789, 8005) verify the propagation plumbing, which is good, but the real scriptjudge.LLMJudge almost certainly returns a zero Verdict on DeadlineExceeded. So in production the timeout fact will have empty SHA and zero latency/tokens. That's fine as an audit signal ("timeout happened"), just worth being aware that the test is exercising the plumbing rather than reflecting realistic prod values.
4. Worth a quick grep for downstream consumers of the fact outcome enum to make sure dashboards or aggregators recognize the new script_session_judge_timeout value:
grep -r "script_session_judge_" --include='*.go' --include='*.sql' --include='*.tsx'
5. Follow-up nudge: the existing comment on the timeout constant already calls out that a circuit breaker plus per-agent rate limit is a follow-up. This change makes that follow-up more valuable, since the agent is now told to retry on every timeout. If the upstream LLM is sick, that could become a tight loop. Worth tracking as a next step rather than rolling into this PR.

Nice clean change overall. Happy to approve once the message string nit is addressed (or even as-is if you'd rather handle it separately).

devcool20:
> Thanks for picking this up. The framing in the PR description (especially the before/after table) made it really easy to review. The motivation is clear and the fix is well-scoped: a small, surgical addition to one error branch with focused tests around it.
> 
> A few things I appreciated:
> 
> * Ordering of the checks is exactly right. ErrNotConfigured first, then parent ctx, then judge-only deadline, then fall-through. That preserves all the existing semantics and slots the new behavior in cleanly.
> * The errors.Is(err, ...) || judgeCtx.Err() == ... belt-and-suspenders is a nice touch. Protects against a Judge implementation that swallows or wraps the context error in an unexpected way.
> * Test coverage matches the matrix in the PR description: judge timeout, parent canceled, parent timed out, and judge-canceled-while-parent-active. Easy to map each test back to a row in your table.
> * The new script_session_judge_timeout fact is the right call for forensics. Keeps timeouts distinguishable from policy blocks in audit downstream.
> 
> A few small suggestions to consider before merge. None are blockers:
> 
> 1. The "8s" in the deny message will drift if scriptSessionJudgeCallTimeout ever changes. Could be worth interpolating the constant so the user-facing message stays accurate:
>    Reason: "Clawvisor: script-session call refused, LLM intent judge timed out (" + scriptSessionJudgeCallTimeout.String() + " limit reached). Please retry.",
> 2. The time.Sleep(5 * time.Millisecond) in ParentContextTimedOut is fine in practice but technically racy. <-ctx.Done() after the WithTimeout would be deterministic:
>    ctx, cancel := context.WithTimeout(context.Background(), 1*time.Microsecond)
>    defer cancel()
>    <-ctx.Done()
> 3. The forensic field assertions in JudgeTimeout (xyz789, 8005) verify the propagation plumbing, which is good, but the real scriptjudge.LLMJudge almost certainly returns a zero Verdict on DeadlineExceeded. So in production the timeout fact will have empty SHA and zero latency/tokens. That's fine as an audit signal ("timeout happened"), just worth being aware that the test is exercising the plumbing rather than reflecting realistic prod values.
> 4. Worth a quick grep for downstream consumers of the fact outcome enum to make sure dashboards or aggregators recognize the new script_session_judge_timeout value:
>    grep -r "script_session_judge_" --include='_.go' --include='_.sql' --include='*.tsx'
> 5. Follow-up nudge: the existing comment on the timeout constant already calls out that a circuit breaker plus per-agent rate limit is a follow-up. This change makes that follow-up more valuable, since the agent is now told to retry on every timeout. If the upstream LLM is sick, that could become a tight loop. Worth tracking as a next step rather than rolling into this PR.
> 
> Nice clean change overall. Happy to approve once the message string nit is addressed (or even as-is if you'd rather handle it separately).

appreciate the feedback, all the issues are addressed in the latest commit!

ericlevine:
Thanks for the PR!

PR #548, opened 2026-06-11
feat(llmproxy): headless runner inactivity routing & programmatic wait API

devcool20:
## Overview

This PR introduces support for headless and asynchronous runners by implementing two major enhancements to Clawvisor's approval workflow:

1. **Inactivity-Based Task Routing (`surface=auto`)**
   - Automatically routes approval requests to the asynchronous approval queue when the user has been inactive for a configurable period.
   - Enables autonomous agents and background runners to continue operating even when the user is offline.

2. **Programmatic Wait API (`GET /api/control/approvals/{id}/wait`)**
   - Provides a long-polling endpoint that allows headless runners to efficiently wait for human approval decisions.
   - Eliminates the need for continuous polling loops and reduces database load.

Together, these changes allow Clawvisor to support long-running, autonomous, and headless workflows while maintaining human approval controls.

---

# Before vs. After

| Scenario | Before | After |
|-----------|----------|---------|
| **Headless runner requesting approval (`surface=auto`)** | Runners had to wait for interactive inline approval prompts. If the user was offline or inactive, the runner could hang indefinitely or time out waiting for input that would never arrive. | When user inactivity exceeds `inactivity_threshold_seconds`, Clawvisor automatically routes the task to the asynchronous approval queue and notification channels. The runner immediately receives a task ID and wait nonce without requiring an inline interaction. |
| **Runner waiting for approval status** | Headless runners had to repeatedly poll `GET /api/control/tasks/{id}` in a spin-loop, generating unnecessary database reads and server load. | Runners can use `GET /api/control/approvals/{id}/wait?timeout=N`, allowing the server to hold the connection and return immediately when the approval state changes. |
| **Approval status propagation** | State changes were only visible on the next polling iteration, introducing latency and wasted requests. | With `EventHub` enabled, approval state changes immediately wake waiting clients through event-driven notifications. |
| **Fallback operation without EventHub** | Clients were responsible for implementing their own polling behavior. | The server provides a built-in fallback polling loop that checks approval state once per second when `EventHub` is unavailable. |
| **Redis approval lookup in multi-replica deployments** | Resolving a pending approval by ID required scanning all active hold entries, resulting in approximately `O(K × L)` Redis scan overhead under load. | A secondary index (`clawvisor:lite_pending_approval_map:<id>`) enables direct `O(1)` lookups of the associated hold entry. |
| **Deleted or purged tasks while waiting** | Long-poll requests remained blocked until timeout expiration even when the underlying task had already been deleted. | Both event-driven and polling implementations detect `store.ErrNotFound` and immediately return `status: denied`, releasing server and client resources. |
| **User activity awareness** | Inline approval interception had no concept of user presence or inactivity. | User activity timestamps are persisted and evaluated so `surface=auto` can intelligently decide between inline and asynchronous approval routing. |
| **Notification channel integration** | Approval requests depended on the user being present in the active chat session. | Inactive-user approvals automatically fall through to the existing async workflow, including dashboard, Slack, Telegram, and other configured notification channels. |
| **Scalability for long-running agents** | Long-running or autonomous agents were tightly coupled to active chat sessions. | Headless runners can operate independently, submit approval requests, and wait programmatically for human decisions without requiring a live chat connection. |

---

# Detailed Architectural Breakdown & Technical Decisions

### Background

For `surface=auto` to make an intelligent routing decision, Clawvisor must know whether a user is currently active.

The existing `RecentUserTurns` field within `PostprocessConfig` only stores recent message content and does not retain timestamps. As a result, there was no reliable way to determine how long it had been since the user's last interaction.

### Database Conversation Tracking

To support inactivity detection, a new `conversations` table was introduced.

#### Schema Changes

A new table stores:

```sql
last_user_message_at TIMESTAMP
```

SQLite and PostgreSQL migrations were added to support the new schema.

### Activity Recording

When requests pass through the LLM proxy:

1. `HasNewUserMessage` determines whether the request contains a genuine new user turn.
2. If true, the conversation record is created or updated.
3. `last_user_message_at` is refreshed.

This creates a durable source of truth for user activity.

### Surface Interception Logic

When a model requests a task using:

```text
surface=auto
```

`MaybeInterceptInlineTaskDefinition()` evaluates inactivity.

If:

```text
current_time - last_user_message_at > inactivity_threshold_seconds
```

the inline approval flow is bypassed.

Instead:

- The task is persisted normally.
- Notifications are dispatched.
- The request enters the existing asynchronous approval workflow.

---

### New Endpoint

```http
GET /api/control/approvals/{id}/wait
```

### Purpose

Provides a long-polling API allowing headless runners to wait for approval decisions without repeatedly polling task status endpoints.

Supported identifiers:

- Approval IDs (`cv-*`)
- Database Task IDs

### EventHub Integration

When `EventHub` is enabled:

1. The handler subscribes to the user's event channel.
2. `events.WaitFor()` blocks efficiently.
3. Status changes publish events.
4. Waiting clients are awakened immediately.

Benefits:

- Near-instant approval propagation
- Minimal database traffic
- No polling overhead

### Fallback Polling Mode

When `EventHub` is unavailable:

1. A ticker loop checks approval state once per second.
2. The request exits immediately when a terminal state is observed.

Benefits:

- Works in all deployments
- Requires no additional infrastructure
- Preserves API consistency

### Task Purge Unblocking

Previously, deleted tasks could leave waiting clients blocked until timeout.

Both implementations now detect:

```go
store.ErrNotFound
```

and immediately return:

```json
{
  "status": "denied"
}
```

Benefits:

- Faster client recovery
- Reduced resource usage
- No unnecessary long-held connections

---

### Problem

In multi-replica environments, locating a pending approval required scanning active Redis hold records.

Lookup complexity was approximately:

```text
O(K × L)
```

where:

- `K` = active holds
- `L` = list entries

This became increasingly expensive under load.

### Solution

A secondary Redis index was introduced:

```text
clawvisor:lite_pending_approval_map:<id>
```

The mapping stores:

```text
approval_id -> hold_key
```

### Registration

During:

```go
Hold()
```

the secondary mapping is created.

### Retrieval

During:

```go
PeekByID()
```

the lookup becomes:

1. Read mapping key.
2. Resolve hold key.
3. Read only the target hold list.

No scans are required.

### Cleanup

Mappings are removed during:

- `Resolve()`
- `Drop()`

and inherit the same TTL as the associated hold record.

Benefits:

- O(1) approval lookup
- Reduced Redis load
- Better multi-replica scalability

---

# Verification & Test Findings

### Wait API Test Suite

Verified:

- Pending approvals
- Immediate approvals
- Immediate denials
- Timeouts
- Event-driven state transitions

```bash
go test ./internal/api/handlers/ -run TestControlWaitForApproval

# ok  github.com/clawvisor/clawvisor/internal/api/handlers (0.05s)
```

### Full Handler Package Suite

Verified all integration and gateway tests.

```bash
go test ./internal/api/handlers/...

# ok  github.com/clawvisor/clawvisor/internal/api/handlers 64.770s
```

All 17 integration and gateway verification tests completed successfully.

---

### Environment

Validation was performed against live Docker containers using:

- WAL-mode SQLite
- Docker loopback networking
- Real approval flows

### Inactivity Routing Verification

Procedure:

1. Send a normal user request.
2. Wait longer than the configured inactivity threshold.
3. Submit:

```text
run command
```

Result:

- Inline approval was bypassed.
- A task was created asynchronously.
- A `cv-nonce-*` identifier was returned immediately.

### Wait Endpoint Verification

Call:

```http
GET /api/control/approvals/{id}/wait
```

Result:

```json
{
  "status": "pending"
}
```

with:

```http
Retry-After: 1000
```

when unresolved.

### Approval Transition Verification

Procedure:

1. Exchange loopback container token for admin JWT.
2. Approve task via:

```http
POST /api/tasks/{task_id}/approve
```

3. Observe the waiting request.

Result:

The blocked wait request immediately unblocked and returned:

```json
{
  "status": "approved"
}
```

without waiting for timeout expiration.

---

# Summary

This PR introduces first-class support for headless and asynchronous runners by adding:

- Inactivity-aware approval routing
- Long-polling approval wait APIs
- Event-driven approval notifications
- Efficient Redis O(1) approval lookups
- Automatic cleanup handling for deleted tasks
- Durable conversation activity tracking

These changes significantly improve the experience for autonomous agents and background runners while reducing database load, improving approval latency, and enabling approval workflows to function correctly even when users are offline.

### DIFF
--- a/internal/api/handlers/connections.go
+++ b/internal/api/handlers/connections.go
@@ -614,7 +614,7 @@ func (h *ConnectionsHandler) PollStatus(w http.ResponseWriter, r *http.Request) 
 func (h *ConnectionsHandler) waitForConnectionResolution(ctx context.Context, connID, userID string, timeout time.Duration) *store.ConnectionRequest {
 	return events.WaitFor(ctx, h.eventHub, userID, timeout,
 		nil, // any event type
-		func(c context.Context) (*store.ConnectionRequest, bool) {
+		func(c context.Context, _ *events.Event) (*store.ConnectionRequest, bool) {
 			cr, err := h.st.GetConnectionRequest(c, connID)
 			if err != nil {
 				return &store.ConnectionRequest{ID: connID, Status: "pending"}, false
--- a/internal/api/handlers/control.go
+++ b/internal/api/handlers/control.go
@@ -371,6 +371,8 @@ func (h *LLMControlHandler) CheckoutTask(w http.ResponseWriter, r *http.Request)
 	})
 }
 
+var errSentinelHold = &llmproxy.PendingLiteApproval{ID: "err-sentinel"}
+
 func (h *LLMControlHandler) WaitForApproval(w http.ResponseWriter, r *http.Request) {
 	agent := middleware.AgentFromContext(r.Context())
 	if agent == nil {
@@ -418,24 +420,27 @@ func (h *LLMControlHandler) WaitForApproval(w http.ResponseWriter, r *http.Reque
 		taskID = pending.PendingTaskID
 	}
 
-	task, err := h.Store.GetTask(r.Context(), taskID)
-	if err != nil {
-		if errors.Is(err, store.ErrNotFound) {
-			if pending == nil {
-				writeJSON(w, http.StatusNotFound, map[string]any{
-					"error":   "not_found",
-					"message": "approval or task not found",
+	var task *store.Task
+	if pending == nil || pending.PendingTaskID != "" {
+		task, err = h.Store.GetTask(r.Context(), taskID)
+		if err != nil {
+			if errors.Is(err, store.ErrNotFound) {
+				if pending == nil {
+					writeJSON(w, http.StatusNotFound, map[string]any{
+						"error":   "not_found",
+						"message": "approval or task not found",
+					})
+					return
+				}
+				writeJSON(w, http.StatusOK, map[string]any{"status": "denied"})
+				return
+			} else {
+				writeJSON(w, http.StatusInternalServerError, map[string]any{
+					"error":   "internal_error",
+					"message": "failed to lookup task: " + err.Error(),
 				})
 				return
 			}
-			writeJSON(w, http.StatusOK, map[string]any{"status": "denied"})
-			return
-		} else {
-			writeJSON(w, http.StatusInternalServerError, map[string]any{
-				"error":   "internal_error",
-				"message": "failed to lookup task: " + err.Error(),
-			})
-			return
 		}
 	}
 
@@ -462,7 +467,10 @@ func (h *LLMControlHandler) WaitForApproval(w http.ResponseWriter, r *http.Reque
 
 	if h.EventHub != nil {
 		if task != nil {
-			resolvedTask := events.WaitFor(r.Context(), h.EventHub, agent.UserID, timeout, []string{"tasks"}, func(c context.Context) (*store.Task, bool) {
+			resolvedTask := events.WaitFor(r.Context(), h.EventHub, agent.UserID, timeout, []string{"tasks"}, func(c context.Context, evt *events.Event) (*store.Task, bool) {
+				if evt != nil && evt.ID != "" && evt.ID != taskID {
+					return nil, false
+				}
 				t, err := h.Store.GetTask(c, taskID)
 				if err != nil {
 					if errors.Is(err, store.ErrNotFound) {
@@ -480,10 +488,13 @@ func (h *LLMControlHandler) WaitForApproval(w http.ResponseWriter, r *http.Reque
 				return
 			}
 		} else if pending != nil && h.PendingApprovals != nil {
-			resolvedHold := events.WaitFor(r.Context(), h.EventHub, agent.UserID, timeout, []string{"audit", "queue"}, func(c context.Context) (*llmproxy.PendingLiteApproval, bool) {
+			resolvedHold := events.WaitFor(r.Context(), h.EventHub, agent.UserID, timeout, []string{"audit", "queue"}, func(c context.Context, evt *events.Event) (*llmproxy.PendingLiteApproval, bool) {
+				if evt != nil && evt.ID != "" && evt.ID != id && (pending.PendingTaskID == "" || evt.ID != pending.PendingTaskID) {
+					return nil, false
+				}
 				p, err := h.PendingApprovals.PeekByID(c, id)
 				if err != nil {
-					return nil, false
+					return errSentinelHold, false
 				}
 				return p, p == nil
 			})
--- a/internal/api/handlers/control.go
+++ b/internal/api/handlers/control.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"net/http"
@@ -8,6 +9,7 @@ import (
 	"time"
 
 	"github.com/clawvisor/clawvisor/internal/api/middleware"
+	"github.com/clawvisor/clawvisor/internal/events"
 	"github.com/clawvisor/clawvisor/internal/intent"
 	"github.com/clawvisor/clawvisor/internal/runtime/llmproxy"
 	"github.com/clawvisor/clawvisor/pkg/store"
@@ -24,7 +26,9 @@ type LLMControlHandler struct {
 	// that the derived capability (placeholder + host + methods +
 	// path prefixes) is checked against the active task scope. nil
 	// falls back to a deterministic check only.
-	IntentVerifier intent.Verifier
+	IntentVerifier   intent.Verifier
+	PendingApprovals llmproxy.PendingApprovalCache
+	EventHub         events.EventHub
 }
 
 func NewLLMControlHandler(baseURL string) *LLMControlHandler {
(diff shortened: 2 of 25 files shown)


ericlevine:
Thanks for putting this together, @devcool20. The overall shape is solid and the EventHub plus ticker fallback is the right approach for headless wait. Before merging, I want to flag a few correctness and authorization concerns that I think are worth another pass.

1. WaitForApproval can return approved on a transient cache error. In the EventHub branch that handles the pending only case in control.go, the fetch lambda returns nil and false on a PeekByID error. Looking at events WaitFor, when the timeout or context fires it re-runs the fetch and returns whatever value comes back, regardless of the done flag. So a Redis hiccup at the deadline produces resolvedHold equal to nil, which the handler then reports as approved. For a runner that is about to take an action behind that verdict, this is an authorization leak. The fix that the task branch already uses is to return a non nil sentinel on error so the post condition can distinguish really cleared from could not tell. I would mirror that pattern here.
2. The pending only branches in both EventHub and polling modes look unreachable to me. By the time control flow reaches the long poll section, the earlier GetTask call has either returned 404 or denied for ErrNotFound, returned 500 for any other error, or produced a non nil task. So the else if pending is not nil branches never run. Either the early denied path at the top of the handler is too eager and these branches should be reachable for the pending exists but task not created yet case, or the branches are dead code and should be removed. Right now it is ambiguous which one is intended.
3. The early denied shortcut may be premature. If a pending hold exists but PendingTaskID is empty, taskID falls back to the approval id, GetTask returns ErrNotFound, and the handler reports denied even though the approval is live in the cache. I would guard that early return with a PendingTaskID is not empty check, or wait for the pending entry to clear before deciding.
4. The new Redis secondary index puts both approval ids and task ids into the same key namespace. They come from different generators, so if they ever collide one mapping clobbers the other. Splitting into clawvisor lite pending approval map approval and clawvisor lite pending approval map task is cheap to do now while the index is new.
5. InactivityThresholdSeconds equal to zero silently disables surface auto. If any future PostprocessConfig construction site forgets to plumb the value through, time Since is never less than or equal to zero, so isInactive stays true and every surface auto request falls through to async. That is a silent regression rather than a crash. A defensive minimum at the read site, treating zero as the 300 second default, would catch a missed wire up.

Smaller things I noticed while reading:

The fetch lambda in the EventHub task branch fires for every tasks event for the user, not just for the task id we care about, and issues a GetTask each time. On a busy user this is gratuitous load. Filtering by event id inside the lambda would help.

I did not see clamping on the timeout query param. Worth confirming there is an upper bound so clients cannot pin a connection indefinitely.

In the Redis Drop path, the json Unmarshal error is swallowed. If a corrupted entry is in Redis the secondary index cleanup is skipped silently and the index key leaks until TTL. Even just logging the error would help diagnose later.

UpdateConversationActivity runs synchronously on the rn. Under SQLite WAL that is a per conversationserialized write. The consumer does not need sub second resolution, so an in process debounce of one write per N seconds per conversation id would take it off the hot path.

The surviving comment in inline task intercept still ly signal we honor in production. Worth updating nowthat auto is also accepted.

Test wise, the goroutine in the EventHub test case calls t Errorf after the handler may have already returned, which races with the test completing. Use t Logf or coordinate via a channel. Iage for ownership rejection, surface auto with bothactive and inactive activity rows, and a Redis PeekByID round trip.

devcool20:
> Thanks for putting this together, @devcool20. The overall shape is solid and the EventHub plus ticker fallback is the right approach for headless wait. Before merging, I want to flag a few correctness and authorization concerns that I think are worth another pass.
> 
> 1. WaitForApproval can return approved on a transient cache error. In the EventHub branch that handles the pending only case in control.go, the fetch lambda returns nil and false on a PeekByID error. Looking at events WaitFor, when the timeout or context fires it re-runs the fetch and returns whatever value comes back, regardless of the done flag. So a Redis hiccup at the deadline produces resolvedHold equal to nil, which the handler then reports as approved. For a runner that is about to take an action behind that verdict, this is an authorization leak. The fix that the task branch already uses is to return a non nil sentinel on error so the post condition can distinguish really cleared from could not tell. I would mirror that pattern here.
> 2. The pending only branches in both EventHub and polling modes look unreachable to me. By the time control flow reaches the long poll section, the earlier GetTask call has either returned 404 or denied for ErrNotFound, returned 500 for any other error, or produced a non nil task. So the else if pending is not nil branches never run. Either the early denied path at the top of the handler is too eager and these branches should be reachable for the pending exists but task not created yet case, or the branches are dead code and should be removed. Right now it is ambiguous which one is intended.
> 3. The early denied shortcut may be premature. If a pending hold exists but PendingTaskID is empty, taskID falls back to the approval id, GetTask returns ErrNotFound, and the handler reports denied even though the approval is live in the cache. I would guard that early return with a PendingTaskID is not empty check, or wait for the pending entry to clear before deciding.
> 4. The new Redis secondary index puts both approval ids and task ids into the same key namespace. They come from different generators, so if they ever collide one mapping clobbers the other. Splitting into clawvisor lite pending approval map approval and clawvisor lite pending approval map task is cheap to do now while the index is new.
> 5. InactivityThresholdSeconds equal to zero silently disables surface auto. If any future PostprocessConfig construction site forgets to plumb the value through, time Since is never less than or equal to zero, so isInactive stays true and every surface auto request falls through to async. That is a silent regression rather than a crash. A defensive minimum at the read site, treating zero as the 300 second default, would catch a missed wire up.
> 
> Smaller things I noticed while reading:
> 
> The fetch lambda in the EventHub task branch fires for every tasks event for the user, not just for the task id we care about, and issues a GetTask each time. On a busy user this is gratuitous load. Filtering by event id inside the lambda would help.
> 
> I did not see clamping on the timeout query param. Worth confirming there is an upper bound so clients cannot pin a connection indefinitely.
> 
> In the Redis Drop path, the json Unmarshal error is swallowed. If a corrupted entry is in Redis the secondary index cleanup is skipped silently and the index key leaks until TTL. Even just logging the error would help diagnose later.
> 
> UpdateConversationActivity runs synchronously on the rn. Under SQLite WAL that is a per conversationserialized write. The consumer does not need sub second resolution, so an in process debounce of one write per N seconds per conversation id would take it off the hot path.
> 
> The surviving comment in inline task intercept still ly signal we honor in production. Worth updating nowthat auto is also accepted.
> 
> Test wise, the goroutine in the EventHub test case calls t Errorf after the handler may have already returned, which races with the test completing. Use t Logf or coordinate via a channel. Iage for ownership rejection, surface auto with bothactive and inactive activity rows, and a Redis PeekByID round trip.

Hey @ericlevine, I have addressed all your feedback points:

Transient Wait Leakage: Added a sentinel check (errSentinelHold) in WaitForApproval to prevent cache query errors from incorrectly reporting as approved.
Task Lookup Guard: Bypassed early database GetTask lookups when PendingTaskID is empty so that pending-only wait loops remain reachable.
Redis Key Namespaces: Split map namespaces using :approval: and :task: prefixes, branching lookups via cv- prefix checks.
Activity Debouncing & Thresholds: Defaulted the threshold to 300s when <= 0 and wrapped activity updates in a 10s cooldown map debouncer.
Event Filtering: Updated the WaitFor callback to accept an event pointer, matching on IDs to skip database checks for irrelevant publishes.
All handlers and LLM proxy unit tests pass successfully!

devcool20:
@cubic-dev-ai review

PR #550, opened 2026-06-11
test(llmproxy): end-to-end coverage for AskUserQuestion substitution shape in inline task intercept

adityajha2005:
## Summary

- Adds two end-to-end tests for the `AskUserQuestion` picker substitution shape introduced in #545
- `inline_task_intercept_test.go` had zero uses of `ApprovalContext.AvailableTools` — all existing tests ran the text-only fallback path; the `AskUserQuestion` branch was untested through `Postprocess`

## Tests added

**`TestPostprocess_InlineTaskInterceptEmitsAskUserQuestionPickerWhenAvailable`**
Sets `AvailableTools: []string{AskUserQuestionToolName}` and asserts the output contains both a text content block (approval prompt + `[clawvisor:approval=...]` marker) and a `"name":"AskUserQuestion"` tool_use block, with text preceding the picker.

**`TestPostprocess_InlineTaskInterceptEmitsTextOnlyWithoutAskUserQuestion`**
Leaves `AvailableTools` empty and asserts the approval prompt renders as text-only with no `AskUserQuestion` tool_use — regression anchor for agents/model versions that don't declare the tool.

Both tests verify the inner hold is registered at `StageAwaitingTaskApproval` regardless of substitution shape.

## Test plan
- [x] `go test ./internal/runtime/llmproxy/postproc/... -run TestPostprocess_InlineTaskInterceptEmits` — both pass
- [x] `go test ./internal/runtime/llmproxy/...` — full suite green
- [x] `go vet ./...` — clean

### DIFF
--- a/internal/runtime/llmproxy/postproc/inline_task_intercept_test.go
+++ b/internal/runtime/llmproxy/postproc/inline_task_intercept_test.go
@@ -1274,3 +1274,114 @@ func TestPostprocess_InlineTaskMalformedBodyFallsThrough(t *testing.T) {
 		t.Fatalf("expected fallback to regular control rewrite on missing purpose; got %s", got.Body)
 	}
 }
+
+// When AskUserQuestion is in the request's available tools, the intercept
+// must emit both a text block (approval prompt + marker) and a synthetic
+// AskUserQuestion tool_use (the yes/no picker) — the substitution shape
+// introduced by #545.
+func TestPostprocess_InlineTaskInterceptEmitsAskUserQuestionPickerWhenAvailable(t *testing.T) {
+	cache := llmproxy.NewMemoryPendingApprovalCache(time.Minute)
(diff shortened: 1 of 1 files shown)


devcool20 (commented):
@ericlevine, looks good. clean test coverage for both AskUserQuestion and text-only fallback paths, with proper ordering and state validation. ready to merge.

ericlevine:
Thanks for closing this coverage gap, @adityajha2005! The two-shape split (picker-present vs. text-only fallback) is exactly the regression anchor the #545 substitution shape needed — `AvailableTools` was genuinely untested through `Postprocess` before this, and the ordering assertion (`text` before `tool_use`) pins down the documented contract from `inline_task_intercept.go:627-643`. Clean, idiomatic, slots right in next to the existing `inline_task_intercept_test.go` scaffolding. Merging now. 🙏

PR #569, opened 2026-06-18
fix(llmproxy): consolidate drift claim rollback logic and handle auto-approve outcome failure

devcool20:
## Summary

This PR addresses two critical scope-drift lifecycle issues identified in the follow-up review of PR #564 (Scope-Drift Continuation Menu). It consolidates the claim/rollback state machine across interceptors and introduces robust error handling for registry writes during task auto-approval.

---

### 1. Auto-Approve SetOutcome Registry Failures (P1)
* **The Problem**: When the conversation-based auto-approval path successfully executes `CreateInlineApprovedTask`, it immediately calls `SetOutcome(ctx, driftID, ScopeDriftOutcomeSucceeded)`. Previously, the registry write error was ignored (`_ =`), and the interceptor returned a successful continuation payload. If the registry write failed (e.g., due to transient database or storage network errors), the task was created, but the drift registry never recorded a succeeded outcome or pre-clear. On the agent's next retry, the original blocked `tool_use` would fail again with no retry path or user-visible recovery.
* **The Fix**: We now explicitly check the error returned by `SetOutcome`. If it fails, the error is audited and traced, and the interceptor returns `(conversation.ToolUseVerdict{}, false)`. This lets the request fall through gracefully to the manual approval/prompt fallback path, where the deferred guard automatically rolls back the claim so that the drift remains usable and retryable.
* **File Affected**: [inline_task_intercept.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/inline_task_intercept.go)

### 2. Consolidating Scattered Claim/Rollback Logic (P2)
* **The Problem**: The rollback logic—including the setup of `claimedDriftID`, `interceptSuccess`, and context-detaching rollback closures—was duplicate and scattered across three different files (`inline_expansion_intercept.go`, `inline_task_intercept.go`, and `scope_drift_intercept.go`). This duplicate state machine split one lifecycle across multiple boundaries, making future exit paths extremely fragile and easy to get wrong.
* **The Fix**: Centralized the claim/rollback lifecycle into a single struct: `DriftClaimGuard` in [scope_drift_intercept.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/scope_drift_intercept.go).
  * The guard manages checking, validation, success flagging, and context-detaching rollback (using `context.WithoutCancel` with a 5-second timeout).
  * Interceptors now leverage `DriftClaimGuard` via a clean `defer guard.Rollback()` pattern at the entry points:
    ```go
    guard := NewDriftClaimGuard(req.Context(), cfg.ScopeDrifts, parsed.DriftID)
    defer guard.Rollback()
    
    // Claim when needed
    _, claimedOk := guard.Claim(...)
    
    // On happy path
    guard.Success()
    ```
* **Files Affected**:
  * [scope_drift_intercept.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/scope_drift_intercept.go)
  * [inline_expansion_intercept.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/inline_expansion_intercept.go)
  * [inline_task_intercept.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/inline_task_intercept.go)

---

### E2E Tests
* Added a new test `TestScopeDriftE2E_NewTaskAutoApproveSetOutcomeFailureRollsBackDrift` in [scope_drift_e2e_test.go](file:///c:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/scope_drift_e2e_test.go) using a custom `failingOutcomeRegistry` mock. The test asserts that:
  1. If `SetOutcome` fails during auto-approval, the interceptor falls through.
  2. The claim is rolled back and cleared from the registry.
  3. The drift is immediately claimable again.

### Automated Test Runs
Executed the entire `llmproxy` E2E test suite:
```powershell
go test -v ./internal/runtime/llmproxy/... -run TestScopeDrift

<!-- This is an auto-generated description by cubic. -->
<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>
<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/internal/e2e/lite/scope_drift_counters.go
+++ b/internal/e2e/lite/scope_drift_counters.go
@@ -71,6 +71,10 @@ func (c *countingScopeDriftRegistry) SetOutcome(ctx context.Context, driftID str
 	return nil
 }
 
+func (c *countingScopeDriftRegistry) RollbackClaim(ctx context.Context, driftID string) error {
+	return c.inner.RollbackClaim(ctx, driftID)
+}
+
 func (c *countingScopeDriftRegistry) LookupPreClear(ctx context.Context, agentID, fingerprint string) (string, bool) {
 	driftID, ok := c.inner.LookupPreClear(ctx, agentID, fingerprint)
 	if ok {
--- a/internal/runtime/llmproxy/inline_expansion_intercept.go
+++ b/internal/runtime/llmproxy/inline_expansion_intercept.go
@@ -137,6 +137,22 @@ func MaybeInterceptInlineExpansion(
 		return conversation.ToolUseVerdict{}, false
 	}
 
+	guard := NewDriftClaimGuard(req.Context(), cfg.ScopeDrifts, parsed.DriftID)
+	defer guard.Rollback()
+
+	if parsed.DriftID != "" {
+		_, claimedOk := guard.Claim(
+			cfg.AgentID,
+			cfg.ConversationID,
+			ScopeDriftOptionExpand,
+			"",
+			audit,
+		)
+		if !claimedOk {
+			return conversation.ToolUseVerdict{}, false
+		}
+	}
+
 	// Land the pending state in the DB before holding so the dashboard
 	// sees the in-flight expansion as a pending row even while the
 	// chat anchor owns the decide window. The creator runs the same
@@ -263,6 +279,7 @@ func MaybeInterceptInlineExpansion(
 			YesDescription: "Authorize the additional scope",
 		})
 	}
+	guard.Success()
 	return verdict, true
 }
 
--- a/internal/runtime/llmproxy/inline_task_intercept.go
+++ b/internal/runtime/llmproxy/inline_task_intercept.go
@@ -252,6 +252,13 @@ func MaybeInterceptInlineTaskDefinition(
 					if setErr := cfg.ScopeDrifts.SetOutcome(req.Context(), parsed.DriftID, ScopeDriftOutcomeSucceeded); setErr != nil {
 						audit("fallthrough", "auto_approve_set_outcome_failed", setErr.Error())
 						trace("inline_task.auto_approve_set_outcome_failed", "err", setErr.Error())
+						if cfg.Store != nil && created != nil && created.ID != "" {
+							rollbackCtx, cancel := context.WithTimeout(context.WithoutCancel(req.Context()), 5*time.Second)
+							defer cancel()
+							if err := cfg.Store.RevokeTask(rollbackCtx, created.ID, cfg.AgentUserID); err != nil {
+								trace("inline_task.auto_approve_outcome_rollback_failed", "task_id", created.ID, "err", err.Error())
+							}
+						}
 						return conversation.ToolUseVerdict{}, false
 					}
 				}
@@ -392,10 +399,10 @@ func MaybeInterceptInlineTaskDefinition(
 		if pendingCreator != nil && pendingTaskID != "" {
 			rollbackCtx, cancel := context.WithTimeout(context.WithoutCancel(req.Context()), 5*time.Second)
 			expireErr := pendingCreator.ExpireInlineTask(rollbackCtx, pendingTaskID, cfg.AgentUserID)
+			cancel()
 			if expireErr != nil {
 				trace("inline_task.pending_rollback_failed", "task_id", pendingTaskID, "err", expireErr.Error())
 			}
-			cancel()
 		}
 		// fallthrough — see the audit-decision rationale above.
 		audit("fallthrough", "inline_task_hold_failed", holdErr.Error()+"; deferring to dashboard rewrite")
--- a/internal/runtime/llmproxy/inline_task_intercept.go
+++ b/internal/runtime/llmproxy/inline_task_intercept.go
@@ -165,6 +165,22 @@ func MaybeInterceptInlineTaskDefinition(
 		return conversation.ToolUseVerdict{}, false
 	}
 
+	guard := NewDriftClaimGuard(req.Context(), cfg.ScopeDrifts, parsed.DriftID)
+	defer guard.Rollback()
+
+	if parsed.DriftID != "" {
+		_, claimedOk := guard.Claim(
+			cfg.AgentID,
+			cfg.ConversationID,
+			ScopeDriftOptionNewTask,
+			"",
+			audit,
+		)
+		if !claimedOk {
+			return conversation.ToolUseVerdict{}, false
+		}
+	}
+
 	// Risk assessment runs BEFORE the hold so the auto-approval gate
 	// can decide whether to skip the human prompt entirely. The
 	// assessment is also used to render the prompt on the fall-through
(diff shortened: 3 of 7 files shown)


ericlevine:
Thanks for tackling this the direction is right (centralizing the lifecycle and falling through cleanly on SetOutcome failure), and the new cross-agent guard on expand/new_task is real hardening. A few things I'd like resolved before this merges.

1. DriftClaimGuard.Rollback() fires even when Claim() never claimed (critical)

The guard tracks success, but not whether we actually own a claim. The defer is wired before any claim attempt, so every "claim-failed" return ErrDriftAlreadyResolved, cross-agent rejection, ErrDriftNotFound, lookup error â still calls registry.RollbackClaim(driftID) and wipes ChosenOption/Outcome/AgentNote on the live record.

The main consequence is that the one-shot cap silently breaks:

1. expand request claims drift ChosenOption=expand.
2. A sibling one_off against the same D hits ErrDriftAlreadyResolved, the intercept returns false.
3. The deferred rollback fires and clears ChosenOption.
4. The drift is now claimable again while the user is still looking at the expand prompt a third call can race in and resolve under a different option.

TestScopeDriftE2E_OneShotCapExpandAndOneOff only asserts the second intercept returned claimed=false; it never re-Gets the drift to confirm the prior ChosenOption is still expand. Adding that assertion would have caught this, and I'd ask you to add it.

The same shape is also a small cross-session integrity hazard: a request from a different agent/conversation that learned a drift_id gets rejected by the new cross-agent guard (good), but the deferred rollback still clears the legitimate session's in-flight claim.

Suggested fix: add a claimed bool set inside Claim() only when validateAndClaimDrift returns (_, true), and have Rollback() short-circuit unless claimed && !success. While you're in there, please update the doc comment too - it already says "if the claim succeeded but the interceptor exited early," which is the intent but not what the code does today.

2. Auto-approve SetOutcome failure orphans the just-created task and skips the approve audit

In inline_task_intercept.go the order is:

1. CreateInlineApprovedTask lands a real, scope-bearing task in the DB.
2. SetOutcome(Succeeded) is called. On failure the new code returns (verdict{}, false).

Two follow-on problems:

- Orphan / duplicate. The task is already created. After the deferred guard rollback (or once you've fixed that, after explicit cleanup) the agent's retry re-enters the inline path, the auto-approve gate fires again, and a second task gets minted with the same purpose. The previous behavior was "ignore the error, working task, broken retry"; the new behavior is "unblocked retry, possibly two tasks." Worth either expiring the just-created task on this path (this for pendingTaskID when the hold fails - same shape) or at minimum documenting the tradeoff.
- Audit gap. audit("approve", "auto_approved_from_coneTaskAutoApproved are below the early return, so theonly audit row left behind is fallthrough auto_approve_set_outcome_failed. An operator triaging "where did this task come from?" won't
find it. Either log the approve before returning, or thing to find.

3. Unrelated cancel() reorder

In inline_task_intercept.go, the cancel() after the r from before to after the if expireErr block.Functionally equivalent here, but unrelated to the PR's stated goals and adds noise to the diff. Please drop it or split it out.

Smaller things

- validateAndClaimDrift is only called by DriftClaimGuard.Claim. Consider folding it into the method so the guard is self-contained.
- The new tests are a real upgrade and cover the happne gap is the "rejected sibling call doesn't clobberthe prior claim" assertion called out in (1).

Happy to take another look once the guard semantics are tightened up - the rest is solid.

devcool20:
> Thanks for tackling this the direction is right (centralizing the lifecycle and falling through cleanly on SetOutcome failure), and the new cross-agent guard on expand/new_task is real hardening. A few things I'd like resolved before this merges.
> 
> 1. DriftClaimGuard.Rollback() fires even when Claim() never claimed (critical)
> 
> The guard tracks success, but not whether we actually own a claim. The defer is wired before any claim attempt, so every "claim-failed" return ErrDriftAlreadyResolved, cross-agent rejection, ErrDriftNotFound, lookup error â�� still calls registry.RollbackClaim(driftID) and wipes ChosenOption/Outcome/AgentNote on the live record.
> 
> The main consequence is that the one-shot cap silently breaks:
> 
> 1. expand request claims drift ChosenOption=expand.
> 2. A sibling one_off against the same D hits ErrDriftAlreadyResolved, the intercept returns false.
> 3. The deferred rollback fires and clears ChosenOption.
> 4. The drift is now claimable again while the user is still looking at the expand prompt a third call can race in and resolve under a different option.
> 
> TestScopeDriftE2E_OneShotCapExpandAndOneOff only asserts the second intercept returned claimed=false; it never re-Gets the drift to confirm the prior ChosenOption is still expand. Adding that assertion would have caught this, and I'd ask you to add it.
> 
> The same shape is also a small cross-session integrity hazard: a request from a different agent/conversation that learned a drift_id gets rejected by the new cross-agent guard (good), but the deferred rollback still clears the legitimate session's in-flight claim.
> 
> Suggested fix: add a claimed bool set inside Claim() only when validateAndClaimDrift returns (_, true), and have Rollback() short-circuit unless claimed && !success. While you're in there, please update the doc comment too - it already says "if the claim succeeded but the interceptor exited early," which is the intent but not what the code does today.
> 
> 2. Auto-approve SetOutcome failure orphans the just-created task and skips the approve audit
> 
> In inline_task_intercept.go the order is:
> 
> 1. CreateInlineApprovedTask lands a real, scope-bearing task in the DB.
> 2. SetOutcome(Succeeded) is called. On failure the new code returns (verdict{}, false).
> 
> Two follow-on problems:
> 
> * Orphan / duplicate. The task is already created. After the deferred guard rollback (or once you've fixed that, after explicit cleanup) the agent's retry re-enters the inline path, the auto-approve gate fires again, and a second task gets minted with the same purpose. The previous behavior was "ignore the error, working task, broken retry"; the new behavior is "unblocked retry, possibly two tasks." Worth either expiring the just-created task on this path (this for pendingTaskID when the hold fails - same shape) or at minimum documenting the tradeoff.
> * Audit gap. audit("approve", "auto_approved_from_coneTaskAutoApproved are below the early return, so theonly audit row left behind is fallthrough auto_approve_set_outcome_failed. An operator triaging "where did this task come from?" won't
>   find it. Either log the approve before returning, or thing to find.
> 
> 3. Unrelated cancel() reorder
> 
> In inline_task_intercept.go, the cancel() after the r from before to after the if expireErr block.Functionally equivalent here, but unrelated to the PR's stated goals and adds noise to the diff. Please drop it or split it out.
> 
> Smaller things
> 
> * validateAndClaimDrift is only called by DriftClaimGuard.Claim. Consider folding it into the method so the guard is self-contained.
> * The new tests are a real upgrade and cover the happne gap is the "rejected sibling call doesn't clobberthe prior claim" assertion called out in (1).
> 
> Happy to take another look once the guard semantics are tightened up - the rest is solid.

Hey @ericlevine, thanks for the thorough review! I've addressed all the feedback points and pushed the updates to the branch:

• DriftClaimGuard.Rollback() firing when Claim() never claimed:

Fix: Added a claimed boolean field to DriftClaimGuard. It is only set to true inside Claim() once registry.ClaimOption returns successfully. Rollback() now checks !claimed at the entry point and returns early.
Verification: Added reg.Get(ctx, drift.ID) and asserted that ChosenOption remains expand at the end of TestScopeDriftE2E_OneShotCapExpandAndOneOff (verifying that the subsequent failed claim does not wipe the active claim's state).

• Auto-approve SetOutcome failure task orphaning:

Fix: If SetOutcome fails in the auto-approve branch of MaybeInterceptInlineTaskDefinition, we now invoke cfg.Store.RevokeTask using a detached context (context.WithoutCancel with a 5-second timeout) to cleanup the just-created active task so it does not get orphaned.

• Diff noise with cancel() reordering:

Fix: Reverted the positioning of cancel() in the pending task hold rollback logic in inline_task_intercept.go to occur immediately after ExpireInlineTask returns, matching the original code structure and removing unrelated changes from the diff.

• Folding validateAndClaimDrift:

Fix: Folded all the validation and claim logic from validateAndClaimDrift directly into DriftClaimGuard.Claim(), removing the unnecessary layer of indirection.

ericlevine:

PR #584, opened 2026-06-25
fix(tasks): address task completion edge cases, expiration races, and bodyless curl rewrites

devcool20:
# Summary

This pull request resolves a series of critical, major, and minor concurrency, correctness, and authentication bugs in the task completion, expiration, and LLM command rewrite systems introduced in PR #575 feat(tasks): expose POST /api/control/tasks/{id}/complete to LLM-proxy agents

---

### 1. `[CRITICAL]` Task Status Overwrite Race in Expiry Sweepers
* **Files**: [approvals.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/approvals.go#L687) and [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/control.go#L252)
* **Problem**: The background task sweeper was executing unconditional status updates to mark expired tasks. This created a race condition where a task concurrently completed by the agent or revoked by the user could be overwritten by the sweeper, transitioning a terminal `completed`/`revoked` status back to `expired`.
* **Fix**: Replaced the unconditional status update with an atomic Compare-And-Swap (CAS) transition using `UpdateTaskStatusFrom(ctx, id, "active", "expired")`. If the task has already transitioned out of `active` to a terminal state, the CAS update safely fails and prevents clobbering.

### 2. `[CRITICAL]` Silent Database Error Swallowing in CAS Retry Loop
* **File**: [tasks.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/tasks.go#L2067-L2070)
* **Problem**: In the event of a CAS failure on task completion, the handler attempts to re-read the task and retry the update. If this retry database update failed, the error was silently swallowed, leaving the client with an incorrect assumption that the completion succeeded or leaving them in an unhandled state.
* **Fix**: Explicitly check the error from `UpdateTaskStatusFrom` in the retry block and propagate it up, returning a `500 Internal Server Error` to the caller.

### 3. `[MAJOR]` Bodyless Completion Curls Failing Authentication
* **Files**: [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/controltool/control.go#L731-L733), [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/controltool/control.go#L906-L908), and [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/controltool/control.go#L1293-L1307)
* **Problem**: The AI model occasionally issues bodyless commands like `curl -sS https://clawvisor.local/control/tasks/task-abc/complete`. Standard `curl` defaults to `GET` for bodyless calls. However, `/complete` is a `POST` endpoint, so the daemon minted a `POST` nonce but the curl call arrived as a `GET` request. This triggered a `403 NONCE_TARGET_MISMATCH` (method mismatch).
* **Fix**: 
  - Updated `controlMethodForCall` to default `/complete` path matching to `POST` even when no body payload is detected.
  - Modified the structured tool rewriter to coerce `method` to `POST`.
  - Added `-X POST` injection to rewritten command strings if the target synthetic endpoint requires `POST` and no explicit method/data flags exist in the command.

### 4. `[MAJOR]` Task Deletion during CAS Race returning HTTP `409`
* **File**: [tasks.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/tasks.go#L2047-L2063)
* **Problem**: If a task was concurrently deleted by the user while the agent was attempting to complete it, the failed CAS update re-read the task and couldn't find it. Rather than informing the agent of the deletion, it fell back to returning `409 Conflict`.
* **Fix**: Added explicit `store.ErrNotFound` and `nil` checks on the re-read. If the task is missing, the handler now correctly returns a `404 Not Found` response.

### 5. `[MAJOR]` Leaked Chain Facts for Expired Tasks
* **Files**: [approvals.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/approvals.go#L697-L699), [tasks.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/tasks.go#L1118-L1120), and [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/control.go#L254-L256)
* **Problem**: When tasks expired naturally or via background sweeps, their associated chain context facts (used for LLM context tracking) were left in the database, leading to resource leakages.
* **Fix**: Integrated a cleanup step that calls `DeleteChainFactsByTask` immediately after any successful transition of a task to `expired`.

### 6. `[MAJOR]` Missing Webhook Callbacks on Task Completion
* **File**: [tasks.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/tasks.go#L2087-L2095)
* **Problem**: The orchestrator was not notified of task completions if a callback URL was configured, as the webhook callback dispatcher was missing from the `Complete` handler.
* **Fix**: Added `dispatchCallback` to dispatch signed completion payloads to the agent's callback URL.

### 7. `[MINOR]` Trailing Slashes in Synthetic Completion URLs
* **File**: [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/runtime/llmproxy/controltool/control.go#L1221-L1238)
* **Problem**: Commands featuring a trailing slash (e.g. `/complete/`) bypassed normal synthetic URL checks and failed path matching.
* **Fix**: Trims trailing slashes from the path string during synthetic URL normalization.

### 8. `[MINOR]` Missing Logging for Opportunistic Task Expirations
* **Files**: [tasks.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/tasks.go#L1115-L1117) and [control.go](file:///C:/Users/sharm/OneDrive/Documents/projects/clawvisor/internal/api/handlers/control.go#L252-L254)
* **Problem**: List-level task listings perform best-effort opportunistic status flips (active -> expired) to offload the background sweeper. However, database errors during these opportunistic flips were silently swallowed without logging.
* **Fix**: Added `DebugContext` logs for opportunistic status update failures and `WarnContext` logs for fact cleanup failures.

---

## Verification & Test Plan

All changes have been successfully verified through automated testing and reviewed by specialized verification agents.

### Automated Tests Run
The following test suites have been run and passed:
1. **Control Command Rewriting Tests**:
   - `TestRewriteControlToolUse_BodylessCompleteCurlInjectsMethodAndCallerAuth`: Asserts bodyless curls to `/complete` correctly inject `-X POST` and authorization headers.
   - `TestRewriteControlToolUse_TrailingSlashCompleteCurlNormalizes`: Asserts trailing slashes in completion curls are trimmed and routed correctly.
2. **Handlers & Routing Tests**:
   - Verification of the `Complete` route, state concurrency checks, webhook notifications, and CAS updates.

**Command to run tests**:
```bash
go test ./internal/runtime/llmproxy/controltool/... ./internal/api/handlers/...

<!-- This is an auto-generated description by cubic. -->
<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>
<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/internal/api/handlers/approvals.go
+++ b/internal/api/handlers/approvals.go
@@ -684,14 +684,26 @@ func (h *ApprovalsHandler) expireTimedOut(ctx context.Context) {
 		return
 	}
 	for _, task := range expiredTasks {
-		_ = h.st.UpdateTaskStatus(ctx, task.ID, "expired")
+		won, statusErr := h.st.UpdateTaskStatusFrom(ctx, task.ID, "active", "expired")
+		if statusErr != nil {
+			h.logger.WarnContext(ctx, "task expiry cleanup: status flip failed", "task_id", task.ID, "err", statusErr)
+			continue
+		}
+		if !won {
+			// Lost the CAS: task was concurrently completed or revoked. Skip.
+			continue
+		}
+
+		if err := h.st.DeleteChainFactsByTask(ctx, task.ID); err != nil {
+			h.logger.WarnContext(ctx, "chain facts cleanup failed on task expiration (Sweep)", "err", err, "task_id", task.ID)
+		}
 
 		h.updateNotificationMsg(ctx, "task", task.ID, task.UserID, "⏰ <b>Task expired</b>")
 
 		if task.CallbackURL != nil && *task.CallbackURL != "" {
 			cbKey, _ := h.st.GetAgentCallbackSecret(ctx, task.AgentID)
 			// See processExpiredApproval — same reason for using the
-			// bounded dispatcher instead of synchronous DeliverResult.
+			// Bounded dispatcher instead of synchronous DeliverResult.
 			h.dispatchCallback(*task.CallbackURL, &callback.Payload{
 				Type:   "task",
 				TaskID: task.ID,
--- a/internal/api/handlers/approvals.go
+++ b/internal/api/handlers/approvals.go
@@ -703,7 +703,7 @@ func (h *ApprovalsHandler) expireTimedOut(ctx context.Context) {
 		if task.CallbackURL != nil && *task.CallbackURL != "" {
 			cbKey, _ := h.st.GetAgentCallbackSecret(ctx, task.AgentID)
 			// See processExpiredApproval — same reason for using the
-			// Bounded dispatcher instead of synchronous DeliverResult.
+			// bounded dispatcher instead of synchronous DeliverResult.
 			h.dispatchCallback(*task.CallbackURL, &callback.Payload{
 				Type:   "task",
 				TaskID: task.ID,
--- a/internal/api/handlers/control.go
+++ b/internal/api/handlers/control.go
@@ -249,7 +249,10 @@ func (h *LLMControlHandler) ListTasks(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 		if task.ExpiresAt != nil && task.ExpiresAt.Before(now) {
-			_ = h.Store.UpdateTaskStatus(r.Context(), task.ID, "expired")
+			won, statusErr := h.Store.UpdateTaskStatusFrom(r.Context(), task.ID, "active", "expired")
+			if statusErr == nil && won {
+				_ = h.Store.DeleteChainFactsByTask(r.Context(), task.ID)
+			}
 			continue
 		}
 		checkedOut := task.ID == checkoutID
--- a/internal/api/handlers/control.go
+++ b/internal/api/handlers/control.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"encoding/json"
 	"errors"
+	"log/slog"
 	"net/http"
 	"strings"
 	"time"
@@ -250,8 +251,12 @@ func (h *LLMControlHandler) ListTasks(w http.ResponseWriter, r *http.Request) {
 		}
 		if task.ExpiresAt != nil && task.ExpiresAt.Before(now) {
 			won, statusErr := h.Store.UpdateTaskStatusFrom(r.Context(), task.ID, "active", "expired")
-			if statusErr == nil && won {
-				_ = h.Store.DeleteChainFactsByTask(r.Context(), task.ID)
+			if statusErr != nil {
+				slog.DebugContext(r.Context(), "opportunistic task expiration update failed (control List)", "err", statusErr, "task_id", task.ID)
+			} else if won {
+				if err := h.Store.DeleteChainFactsByTask(r.Context(), task.ID); err != nil {
+					slog.WarnContext(r.Context(), "chain facts cleanup failed on task expiration (control List)", "err", err, "task_id", task.ID)
+				}
 			}
 			continue
 		}
--- a/internal/api/handlers/tasks.go
+++ b/internal/api/handlers/tasks.go
@@ -1112,7 +1112,9 @@ func (h *TasksHandler) List(w http.ResponseWriter, r *http.Request) {
 			// Opportunistically mark expired tasks in the DB so the
 			// background sweep doesn't have to catch them later.
 			won, statusErr := h.st.UpdateTaskStatusFrom(ctx, t.ID, "active", "expired")
-			if statusErr == nil && won {
+			if statusErr != nil {
+				h.logger.DebugContext(ctx, "opportunistic task expiration update failed (List)", "err", statusErr, "task_id", t.ID)
+			} else if won {
 				if err := h.st.DeleteChainFactsByTask(ctx, t.ID); err != nil {
 					h.logger.WarnContext(ctx, "chain facts cleanup failed on task expiration (List)", "err", err, "task_id", t.ID)
 				}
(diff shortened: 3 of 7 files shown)


ericlevine:
Thanks for the thorough writeup and for splitting this into clearly categorized fixes. The CAS replacement in the sweepers, the 404-vs-409 branching in the Complete retry, and the trailing-slash normalization all look good. A few things before we land it.

Blocking: rethink the method-coercion approach

The -X POST injection in the command-string rewriter and the raw["method"] = "POST" write in the structured rewriter both make me want to push back. I'd rather see this routed through the failure rewriter than fixed by coercion.

I checked the rest of the implementation. Pre-PR, the rewriter never touches method anywhere. The structured rewriter only modifies url and headers. The command-string rewriter only injects the target-host and caller-auth headers and rewrites the URL. Method has always been trusted from the agent.

That works for every other endpoint because of a property of the existing surface: of the POST endpoints controlMethodForCall recognizes, every one except /complete requires a body. POST /tasks, POST /tasks/{id}/expand, POST /task/checkout, POST /control/autovault/script all carry JSON, and curl auto-promotes a -d/--data call to POST. So for every other POST endpoint, the body itself forces the right method. The agent could omit -X POST and they would all still work.

/complete is the only bodyless POST in the entire surface. It is the only endpoint where the agent has to remember -X POST as a standalone token. That asymmetry is real, but I do not think the right response is to make /complete the only endpoint with method coercion. The existing architecture says "agent emits the right method, daemon enforces mismatch at nonce verification" and the pre-existing comment at control.go:1272-1276 documents that exact failure mode as the intended outcome of forgetting -X POST.

A few specific concerns with the coercion approach:

1. It silently overrides agent intent. The intent verifier reads what the agent emitted. Coercing GET into POST in the rewriter means the verifier rates intent against a call the agent did not actually make. That erodes the audit boundary the proxy is supposed to enforce.
2. It hides agent confusion we would want to see. A clear 403 with "method mismatch: /complete requires POST" is a one-shot correction the model handles trivially on retry. Silent fixes mean the wrong pattern persists in transcripts.
3. The rule duplicates across three places. controlMethodForCall already declares "/complete is POST" as canonical truth. This PR adds the same hardcoded /complete suffix check in two more spots. Anyone adding a future synthetic-POST endpoint has to remember all three. One source of truth is better.

The existing RewriteControlFailureToolUse (control.go:745) is already wired from policies/control_tool_use_evaluator.go:176 for exactly this shape: agent emitted something the proxy cannot route, rewrite the tool use to call /api/control/failure with a structured reason. Detecting that the agent's method does not match the verdict's canonical method and routing through that path gives us one canonical rule, one enforcement point, and an audit-clean error the agent can react to. It also drops both the -X POST injection and the raw["method"] = "POST" write.

Trade-off to flag: this costs one wasted tool call when the agent forgets -X POST. The PR description says this is occasional model behavior, so the agent self-corrects fine. My read is the cleaner architecture is worth that cost, but worth checking with the PR #575 author.

Blocking: client-visible changes worth calling out in the PR description

4. The 409 to 404 change in Complete on concurrent delete is a client-visible status code change. It is the right semantics, but please flag it so reviewers can confirm no downstream orchestrator branches on the oldÂ 409.
5. The new callback dispatch in Complete. Any orchestrator that did not previously expect a "completed" callback will start getting one. Probably fine, but please confirm.

Non-blocking, please fix in this PR

6. approvals.go:693 capitalizes "Bounded" mid-sentence in a comment you did not otherwise touch. Looks like an accidental edit, please revert.

Optional, future

7. control.go uses package-level slog.DebugContext for the new opportunistic-expiry log while tasks.go uses h.logger.DebugContext. That is because LLMControlHandler does not carry a logger field today. Worth a followup when it does.
8. The opportunistic expiry path now runs DeleteChainFactsByTask inside a List request goroutine. Fine for normal traffic, but if ListExpiredTasks ever returns a large backlog this serialises N deletes on the request path. Worth keeping an eye on.

Nice work on the rest of the diff.

devcool20:
@ericlevine, thanks for the catch on method coercion. i dropped the -x post injection and structured post rewrites entirely. instead, method mismatches (like bodyless complete curls) now route to /api/control/failure so the agent fails closed and self-corrects. 

also aligned the callback context timeout in tasks.go to 30s and reverted the capitalized comment in approvals.go.

PR #585, opened 2026-06-25
fix(reliability): keep audit writes alive on client cancel; stop polling on lost telegram lock

ericlevine:
## Summary

Two fixes for the most chronic firing alerts in `clawvisor-production` (last 24h: 38 telegram poller conflicts + 20 audit-log write failures).

* **`LogAudit` (postgres + sqlite)** — detach from the request `ctx` with `context.WithoutCancel`, capped at a 5s timeout. A client disconnect or request-deadline cancellation no longer drops the audit row mid-`INSERT`. Every failing case in the last 24h logged `err="context canceled"`.
* **`PollingLock.Renew` now returns `bool`** so the caller can detect when the lock has been taken by another instance. `pollForCallbacks` exits cleanly in that case instead of issuing another `getUpdates`, which would cause Telegram to terminate the newer consumer with a `Conflict` error and flap polling between instances. The 35s lock TTL is preserved so failover after an instance crash remains fast (the earlier version of this PR bumped TTL to 90s — code review caught that this traded conflict noise for stale-lock latency, so the fix moved to the caller instead).

Neither change is a structural fix — they just stop these two alerts being the dominant noise so the real signal (Cloud Run long-poll 503s, which actually account for the misnomered `Container dying` alert) is easier to see.

## Test plan

- [ ] `go vet ./pkg/store/postgres/... ./pkg/store/sqlite/... ./internal/notify/telegram/...` (clean locally)
- [ ] `go build ./...` (clean locally)
- [ ] After merge + deploy, watch `Audit log write failed` and `Telegram poller multi-instance conflict` alert frequency for 24h
- [ ] Confirm no regression in audit row throughput (audit writes still succeed under normal request flow)

### DIFF
--- a/internal/notify/telegram/polling.go
+++ b/internal/notify/telegram/polling.go
@@ -195,8 +195,15 @@ func (n *Notifier) pollForCallbacks(ctx context.Context, ps *pollingSession) {
 		default:
 		}
 
-		// Renew the distributed lock before each long-poll.
-		n.pollingLock.Renew(ctx, ps.userID)
+		// Renew the distributed lock before each long-poll. If a different
+		// instance has taken over (our TTL expired and the other side
+		// acquired it), exit cleanly here. If we call getUpdates anyway,
+		// Telegram will terminate the newer consumer with a "Conflict"
+		// error and the user's polling will flap between instances.
+		if !n.pollingLock.Renew(ctx, ps.userID) {
+			logger.WarnContext(ctx, "callback polling: lock lost, another instance is polling", "user_id", ps.userID)
+			return
+		}
 
 		updates, err := n.getUpdates(ctx, ps.botToken, offset, 10)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


ericlevine:
Good catch — bumping TTL traded conflict noise for stale-lock latency, especially bad for persistent group observation where `BootstrapGroupObservation` only runs once at startup.

Force-pushed `55e3042` with a more targeted fix:

- Reverted `pollingLockTTL` back to 35s.
- `PollingLock.Renew` now returns `bool` (the Lua script already knew whether we still owned the lock — just propagating that signal).
- `pollForCallbacks` exits cleanly when `Renew` reports the lock was lost, so we never call `getUpdates` after losing ownership. That kills the conflict at its source: Telegram never has to terminate two competing consumers.
- Transient Redis errors are treated as "assume we still own it" so a network blip doesn't cause unnecessary failover.

Failover after an instance crash stays at the original ~35s. Audit-log changes are unchanged.

PR #595, opened 2026-06-27
fix(llmproxy): enforce strict per-conversation task-scope isolation

ericlevine:
## Summary

- Eliminates cross-conversation task-scope leakage. When a conversation has a checked-out task, tool calls authorize **only** against that task — no silent fallback to other active tasks owned by the same agent.
- Closes the secondary leak from empty-`ConversationID` writes by killing the legacy `(user, agent, "")` bucket; `/control/task/checkout` now scopes writes via a proxy-injected `X-Clawvisor-Conversation-ID` header.
- Adds `preferred_task_id` + `task_scope_path` (`preferred_strict` / `no_preferred_fallback` / `preferred_mismatch_blocked`) to `params_safe` so per-conversation isolation is queryable in audit.

## Key changes

| Layer | Change |
|---|---|
| `policy/task_match.go`, `request_classify.go` | Strict preferred-task matching. A non-empty `preferredTaskID` restricts the search to that task; mismatch returns `nil` / `NeedsNewTask` rather than scanning siblings. Stale-checkout pointer also surfaces as `NeedsNewTask` (cubic loop 1 fix). |
| `pkg/runtime/decision/decision.go` | New `SourceTaskScopeMismatchPreferred` so audit can distinguish "preferred mismatched" from "no preferred at all." |
| `llmproxy/taskscope.go` | `TaskScopeChecker.Check` extended with `preferredTaskID`; threaded through legacy fallback in `pipelineeval/factory.go`. |
| `llmproxy/taskcheckout/` | `ErrConversationIDRequired`; `Set` rejects empty CID, `Get` returns not-found (no shared bucket). |
| `inspector/rewriter.go`, `controltool/control.go` | Inject `X-Clawvisor-Conversation-ID` on rewritten control calls. |
| `handlers/control.go` | `/control/task/checkout` requires the injected header; `ListTasks` reads it. |
| `handlers/llm_endpoint.go` | Drop legacy fallback in `checkedOutTaskID`. |
| `conversation/AuditEvent`, `llmproxy/audit.go` | New `params_safe` fields. |

## Test plan

- [x] `go test ./internal/runtime/policy/... ./internal/runtime/llmproxy/... ./pkg/runtime/decision/... ./internal/api/handlers/... ./internal/runtime/conversation/...` — all green
- [x] `go build ./...` clean
- [x] `make test-e2e` (smoke) — 41 PASS / 1 SKIP / 0 FAIL
- [x] `internal/e2e/lite/` LLM scenario `script_session_scope_mismatch_recovery/claude` — passed (was failing during dev with a runaway drift loop; reverting an over-eager `AppliesToSource` addition fixed it)
- [x] Cubic review against `origin/main` — loop 1 caught a stale-checkout leak window (fixed in second commit), loop 2 clean
- [ ] Post-merge: run the brief's verification SQL on staging audit logs (`task_scope_path` distribution + cross-task-per-conversation query)

## Notes

- Brief: `/tmp/clawvisor-fix-task-scope-leak.md`
- One known-flaky lite e2e scenario (`script_session_inline_fanout/claude`) still over scope-mismatch budget after the fix, but that's claude struggling with the proxy URL (its own self-diagnosis in the trace); not caused by this change.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/api/handlers/control.go
+++ b/internal/api/handlers/control.go
@@ -10,6 +10,7 @@ import (
 	"github.com/clawvisor/clawvisor/internal/api/middleware"
 	"github.com/clawvisor/clawvisor/internal/intent"
 	"github.com/clawvisor/clawvisor/internal/runtime/llmproxy"
+	"github.com/clawvisor/clawvisor/internal/runtime/llmproxy/inspector"
 	"github.com/clawvisor/clawvisor/pkg/store"
 	"github.com/google/uuid"
 )
@@ -232,8 +233,15 @@ func (h *LLMControlHandler) ListTasks(w http.ResponseWriter, r *http.Request) {
 
 	checkoutID := ""
 	checkoutUnavailable := false
-	key := llmproxy.TaskCheckoutKey{UserID: agent.UserID, AgentID: agent.ID}
-	if h.TaskCheckouts != nil {
+	// The lite-proxy rewriter injects X-Clawvisor-Conversation-ID on
+	// rewritten control calls so the per-conversation checkout bucket
+	// can be reached. A missing header means the request didn't
+	// originate from an inference turn — there's no scoped bucket to
+	// consult, so we report "no checkout" rather than fall back to the
+	// shared legacy bucket that was the cross-conversation leak source.
+	conversationID := strings.TrimSpace(r.Header.Get(inspector.ConversationIDHeader))
+	key := llmproxy.TaskCheckoutKey{UserID: agent.UserID, AgentID: agent.ID, ConversationID: conversationID}
+	if h.TaskCheckouts != nil && conversationID != "" {
 		if checkout, ok, err := h.TaskCheckouts.Get(r.Context(), key); err != nil {
 			checkoutUnavailable = true
 		} else if ok {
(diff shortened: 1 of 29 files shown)


ericlevine:
## Cubic loop 3 — pre-existing P1 flagged, deferred

Loop 3 of `cubic review --base origin/main` surfaced one P1:

> **Alias-matched task scopes skip intent verification** — `policy.matchTaskScope` allows alias fallback (`gmail` covers a task scoped `gmail:personal`), but `decision.findMatchingAction` only checks exact `a.Service == serviceID`. The aliased match yields `action == nil`, and `runIntentVerify` short-circuits to allow when `action` is nil — bypassing the verifier entirely.

`git blame` on `findMatchingAction` (`pkg/runtime/decision/decision.go`) shows it was added in `bd90ae453` (2026-05-19), more than a month before this PR. **Not introduced by these commits**, and out of scope for the per-conversation isolation invariant this PR enforces.

Calling out separately so it doesn't get lost — worth a focused follow-up PR (likely: fold alias handling into `findMatchingAction`, or wrap the lookup so the alias semantics are consistent across both the matcher and the action resolver).

Loops 1 and 2 were green (loop 1's P1 was fixed in 4c89ca34); this PR's scope-leak changes have no remaining cubic findings of their own.

PR #599, opened 2026-06-30
fix(conversation): first-user-message fingerprint fallback for legacy clients

ericlevine:
## Summary

- Follow-up to #595: close the per-conversation isolation gap for clients that don't carry a native session id on the wire (raw Anthropic API without `metadata.user_id.session_id`, generic OpenAI Responses without `prompt_cache_key`).
- `ConversationID()` now falls back to a SHA-256 fingerprint of the first user message for Anthropic and OpenAI Responses — mirroring the `openAIChatFingerprint` path that already existed for `/v1/chat/completions`.
- Empty CID is now a hard signal of "no conversation exists at all" (no user turn in the body). Callers can rely on a non-empty id whenever there's any user-authored content.

## Why this is needed

The doc on `ConversationID()` previously said:

> Returns "" when no identifier can be derived. Callers must treat empty as "unknown conversation" and fall back to the pre-conversation-scoping behavior — empty IDs MUST collide rather than partition, otherwise old clients silently lose their previous-turn state on every request.

#595 made the taskcheckout strict (rejects empty CID), which closed the leak for Claude Code + Codex but broke checkout persistence for any client whose conversation_id resolved to `""`. The fingerprint fallback restores per-conversation isolation for the full client population without requiring a wire-protocol upgrade.

## Source matrix after this PR

| Provider | Native source | Fallback |
|---|---|---|
| Anthropic `/v1/messages` | `metadata.user_id` → nested `session_id` | **fp-** SHA256(first user message) — NEW |
| OpenAI `/v1/responses` | `prompt_cache_key` | **fp-** SHA256(first user input) — NEW |
| OpenAI `/v1/chat/completions` | Clawvisor minted/echoed marker | `fp-` (already existed) |

`liteProxyConversationIDSource` collapses the per-provider switch into a single `fingerprint` branch — `fp-` is the universal marker.

## Test plan

- [x] `go test ./internal/runtime/conversation/... ./internal/api/handlers/... ./internal/runtime/policy/... ./pkg/runtime/decision/... ./internal/runtime/llmproxy/...` — all green
- [x] New `TestConversationIDAnthropicFingerprintFallback` + `TestConversationIDOpenAIResponsesFingerprintFallback` pin: fp- prefix, distinct-message partitioning, cross-turn stability, block-shape vs string-shape equivalence, native-key precedence.
- [x] `TestConversationIDAnthropicNoSession` updated (was `want empty`, now `want fp-`).
- [x] `TestConversationIDAnthropicMalformed` fixtures narrowed to the cases that legitimately have no user message.
- [x] `TestLLMEndpoint_InlineApprovalReleasesHeldToolUse` + `TestLLMEndpoint_RewritesTaskApprovalReplyBeforeForwarding` now set `metadata.user_id.session_id` so hold + lookup share a conversation (they previously relied on the closed empty-CID collision).
- [x] `TestLLMEndpoint_Anthropic_NoConversationIDMarker` no longer treats `fingerprint` as a Chat-Completions-only label.
- [x] LLM characterization snapshots regenerated.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/599?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ericlevine:
Superseded by the marker+preservation approach in the replacement PR. The fingerprint fallback survives in that PR as the tertiary safety net.

PR #644, opened 2026-07-28
fix(install): honor route in the OpenClaw and Hermes markdown installers

ericlevine:
## Why

Closes the `KNOWN GAP` documented on `installerCtx.Route` in #643.

#643 made `skill-only` the default install route, but only the self-install **shell** targets honoured it. `renderOpenClawInstaller` and `renderHermesInstaller` ignored `ctx.Route` entirely and unconditionally walked the user through pointing their provider at Clawvisor.

OpenClaw is the one that bites: it appears in `LEGACY_AGENT_TABS` (`web/src/pages/Agents.tsx`), the tab set shown to users **without** proxy-lite. So a user with no entitlement could pick OpenClaw, follow a routed doc, and get `403 PROXY_LITE_DISABLED` on their first model call — exactly the failure #643 set out to eliminate. Hermes is proxy-tab-only so gated users never see it, but it carried the identical latent bug.

## Approach

Both renderers now branch into a dedicated skill-only render function, mirroring how the shell targets keep `codex_skill_only.sh.tmpl` beside `codex.sh.tmpl`. Threading `if ctx.Route` through the routed bodies would have put the routed output one careless edit away from breaking; a separate function makes that structurally hard.

Dropped from the skill-only docs: detect-provider, vault-upstream-key, probe-deployment, preflight, and point-at-Clawvisor. Detect-provider and probe go too because their outputs (`$PROVIDER`, `$BASE_PATH`, `$KEY_ENV`, `$OPENCLAW_API`, …) are consumed *only* by the vault and configure steps — keeping them would leave dead variables plus prose telling the user to vault a key "so the proxy can route there."

Three things were also quietly wrong for skill-only and are now route-aware:

- `sectionSelfUninstall` claimed the harness "is now routed through Clawvisor"
- the uninstall recipe listed a provider entry / `model:` block / vaulted key that a skill-only install never creates
- `setupFrontmatter`'s description promised "route every session through Clawvisor by default" — that string is what the user reads in their harness's skill picker

## Verification

**Routed output is byte-identical.** Dumped both targets in all three variants before and after with a fixed `publicURL` (ephemeral httptest ports otherwise mask the comparison) and compared: `openclaw` routed is 19,889 B and `hermes` routed 20,235 B, unchanged. The default drops to 7,998 B / 7,953 B — the skill-only doc. On `main`, default and `route=proxy` were identical, which is the bug in one line.

**Step numbering** is contiguous in all four variants — skill-only 1–4, routed 1–8. A new `assertStepsContiguous` helper regex-extracts `## <n>.` headings and asserts 1..N so this can't silently rot.

**No dangling references** in the skill-only docs: they use only `$AGENT_NAME`, `$CLAWVISOR_APP_URL`, `$TOKEN`, `$TOKEN_FILE`, all assigned in step 1.

**New tests are load-bearing:** swapping in `main`'s `installer.go` fails 3 of the 4 new tests. (`TestInstallerMarkdownProxyRouteOptIn` passes on main by construction, since main always routes.)

Three existing tests asserted routed content via the bare default and now request `route=proxy` explicitly. `go build ./...`, `go vet`, and `go test ./...` all clean; no new `gofmt` complaints (`installer.go` has 6 pre-existing ones, content-identical before and after).

## Note on the dashboard

`Agents.tsx` only appends `route=` for `selfInstall` targets, so the dashboard never sends it for openclaw/hermes and renders no routing toggle for them. That means this fix takes effect for every dashboard-driven markdown install and no UI affordance is lost. `route=proxy` stays reachable by URL and the skill-only doc documents that escape hatch inline. A stale comment there is corrected; the logic is deliberately unchanged. Offering an Observe posture for OpenClaw in the UI would be a separate frontend change.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/api/handlers/installer.go
+++ b/internal/api/handlers/installer.go
@@ -139,16 +139,17 @@ type installerCtx struct {
 	// X-Clawvisor-Agent-Token header while leaving the OAuth session in
 	// Authorization untouched and never referencing an API key.
 	//
-	// KNOWN GAP: only the self-install shell targets (claude-code, codex) honor
-	// this field. The markdown helper renderers — renderOpenClawInstaller and
-	// renderHermesInstaller — ignore it and unconditionally instruct the user to
-	// point their provider at Clawvisor. OpenClaw matters most, because it is in
-	// the dashboard's non-proxy tab set (LEGACY_AGENT_TABS in web/src/pages/
-	// Agents.tsx) and so is reachable by a user who has no proxy-lite
-	// entitlement, who then gets 403 PROXY_LITE_DISABLED. Giving those two a
-	// skill-only variant means dropping their vault-key/preflight/point-at-
-	// Clawvisor steps and renumbering the doc, which is left as follow-up rather
-	// than bundled into the default flip.
+	// Every target honors this field. The self-install shell targets pick a
+	// template (claude_code_skill_only.sh.tmpl / codex_skill_only.sh.tmpl);
+	// the markdown helper renderers pick a render function
+	// (renderOpenClawSkillOnlyInstaller / renderHermesSkillOnlyInstaller),
+	// which drops the detect-provider / vault-key / probe / preflight /
+	// point-at-Clawvisor steps and renumbers the doc from 8 steps to 4. That
+	// matters most for OpenClaw, which is in the dashboard's non-proxy tab set
+	// (LEGACY_AGENT_TABS in web/src/pages/Agents.tsx) and so is reachable by a
+	// user who has no proxy-lite entitlement: while the markdown renderers
+	// ignored this field, such a user got a routed doc and then
+	// 403 PROXY_LITE_DISABLED on their first model call.
 	Route string
 }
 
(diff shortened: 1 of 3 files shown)


ericlevine:
Closing — the premise was wrong, and I should have checked it before opening this.

I justified this PR on the claim that a user without proxy-lite could pick OpenClaw in the dashboard, get a routed install doc, and hit `403 PROXY_LITE_DISABLED`. That is not true. I confirmed `openclaw` is in `LEGACY_AGENT_TABS` and jumped straight to the conclusion, without checking what that tab set actually renders.

It renders a different component (`web/src/pages/Agents.tsx:1052-1057`):

```
proxyLiteUI ? OnePasteGuide        → /skill/install/<target>.md
            : LegacyOpenClawGuide  → a setup prompt, no installer URL
```

`LegacyOpenClawGuide` takes only `setupURL`, emits a natural-language setup prompt, and has zero routing references — it is already a skill/gateway flow and has been working that way for months. `OnePasteGuide`, the only caller of `/skill/install/*.md`, renders **only** when `proxyLiteUI` is true — i.e. for users who *have* proxy-lite, for whom the routed doc is the correct output.

So the markdown renderers ignoring `route` is a real inconsistency but a harmless one: no dashboard path leads an unentitled user to them. Not worth ~470 lines and three newly-authored doc sections.

Note for whoever picks this up: the `KNOWN GAP` comment on `installerCtx.Route` (added in #643) repeats the same incorrect reasoning about OpenClaw being reachable without an entitlement. It should be corrected or dropped.

PR #653, opened 2026-08-07
feat(slack): download file and attachment content

ericlevine:
Slack could list channels and read messages but had no way to reach an uploaded file — despite already requesting the `files:read` scope and advertising "list files" in its description.

## Actions

| Action | How |
|---|---|
| `list_files` / `get_file` | `files.list` / `files.info`, plain YAML |
| `download_file` | Go override |

`download_file` needs Go because the content lives on `files.slack.com` rather than the `slack.com/api` base URL, and returns raw bytes — the YAML REST runtime concatenates `base_url + path` and hard-assumes a JSON body, so neither is expressible. Same `override: go` pattern as Dropbox/Drive/OneDrive.

`list_messages` now also surfaces `files[]`, projected to `id`/`name`/`mimetype`/`size`. The response field whitelist was dropping it, so an agent had no way to learn an attachment existed. The projection matters: a raw Slack file object carries ~40 fields of thumbnail metadata each.

## Always base64

Content comes back base64 for **every** type, including text. A text path can't round-trip a file:

- `format.SanitizeText` strips HTML — an `.html` download would return the prose with the markup deleted — and truncates at 200k runes.
- Even unsanitized, bytes in a JSON string lose anything not valid UTF-8 to U+FFFD, silently corrupting legacy-encoded CSVs and every binary format.

Uniform base64 also makes the caller contract uniform: one `base64 -d` regardless of type.

## Size limits

New `format.MaxDownloadBytes` (10 MB) with a 1 MB default, raisable per-call via `max_bytes`. Deliberately **not** a bump to `MaxBodyLen` — that constant is a rune count used to truncate human-readable text (mail bodies, calendar descriptions) and is applied to every string in every MCP response; downloads had only borrowed it as a byte count.

Oversized files are **refused, not truncated**. A truncated binary is undecodable, and Drive's current behaviour returns it under a `"Downloaded foo.png"` summary — success-shaped, corrupt payload.

## Three Slack-specific hazards

Each found by real-world testing, each with a regression test:

1. **Sign-in page on 200.** An under-scoped token gets HTML with a 200 status. Detection can't just ask "is this HTML?" — that's also true of every uploaded `.html` file, and the first cut rejected all of them. Now requires an attachment disposition or a body length matching `files.info`.
2. **`files.info` scales with file size.** Slack embeds `preview`/`plain_text` for textual files, so a 318 KB HTML file produced >200 KB of metadata; the old read cap truncated the JSON into `unexpected end of JSON input`.
3. **Download URL is response data.** Host is pinned to Slack on the initial request and every redirect hop. Unlike OneDrive's pre-signed URLs, Slack requires the bearer token at the final hop, so it is forwarded deliberately rather than stripped.

## Testing

17 unit tests. Verified separately that the definition passes `yamlvalidate` (a failure drops the *entire* adapter, not just the bad action), that all four actions load, and that `download_file` dispatches to the Go override rather than the REST engine.

**Not covered:** every test uses `httptest`. The `Content-Disposition: attachment` assumption behind sign-in detection comes from Slack's documented behaviour for `url_private_download` and has not been observed against the live API — if Slack omits it, the length check is the fallback, and both must fail together to break a download.

Dropbox, Drive, and OneDrive still have the text branch and the same three corruption paths; aligning them changes their existing output contract, so it's left for a follow-up.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/653?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/adapters/definitions/slack.yaml
+++ b/internal/adapters/definitions/slack.yaml
@@ -92,6 +92,12 @@ actions:
         - { name: text, sanitize: true }
         - { name: ts }
         - { name: thread_ts, optional: true }
+        # Projected down to the fields needed to identify and fetch an
+        # attachment — a raw Slack file object carries ~40 fields of
+        # thumbnail and preview metadata per file.
+        - name: files
+          expr: 'files == nil ? nil : map(files, {{"id": #.id, "name": #.name, "mimetype": #.mimetype, "size": #.size}})'
+          optional: true
       meta:
         - { name: response_metadata.next_cursor, rename: next_cursor }
       summary: "{{len .Data}} message(s)"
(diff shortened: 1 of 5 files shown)


ericlevine:
Thanks — all five findings were valid and are fixed in 63a022bf.

**P1, discarded `io.ReadAll` error** — correct, and it was the same silent-truncation class this PR exists to remove. A connection broken mid-transfer produced a short body that base64-encoded cleanly and returned as a successful download. Now propagated, with the status check still first so a 4xx body surfaces its message rather than being masked by the read failure.

**P2, untested redirect path** — right that this was the riskiest gap: the token-to-final-hop behaviour and host pinning are the security-critical parts and nothing exercised them. Added two tests. The success case asserts the final hop receives `Authorization` and that the redirect target was host-checked, using a recording `hostCheck` rather than a no-op. The negative case redirects to a foreign host and fails if the request arrives there at all, so a leaked token trips the test.

**P2, `maxFileInfoBytes` below the `max_bytes` ceiling** — genuine inconsistency: `files.info` embeds the file's own text, so a large textual file permitted by `max_bytes` could be rejected at the metadata step. Now derived from `format.MaxDownloadBytes` (2× for JSON escaping, plus 1 MiB headroom) so the two limits can't drift apart.

**P2, no HTTP timeouts** — correct, `http.DefaultClient` has none. Added 30s for metadata and 2m for content, the latter sized for a 10 MB transfer.

**P3, stale doc comment** — fixed.

Unrelated: `Browser (Playwright)` failed on `approvals.spec.ts:37`. This PR touches no frontend or approval code, and the identical failure (same test, same assertion, same line) occurs on `main` in run 30654881145. Pre-existing flake.

ericlevine:
Second round addressed in 1ff082e1 — and the redirect finding was a real bug, not just a test-quality issue.

**P2, false confidence in the redirect test.** Correct, and worse than "gives false confidence": it was hiding a genuine defect. Both `httptest` servers ran on `127.0.0.1` differing only by port, and `net/http` strips `Authorization` only when a redirect leaves the origin host — so the header survived on its own.

Verified with a scratch test that a `127.0.0.1` → `localhost` redirect arrives with an empty `Authorization`. Against real Slack, `files.slack.com` redirects to a CDN host that isn't a subdomain of the origin, so `shouldCopyHeaderOnRedirect` would have dropped the token and the final hop would 401. The PR description claimed the token was forwarded deliberately; it wasn't.

`CheckRedirect` now re-applies the header — but only *after* `hostCheck` confirms the target is a Slack host. That ordering is what separates this from a token leak. The test crosses a real hostname boundary and I confirmed it fails without the fix (`token did not survive the redirect: final hop saw ""`).

This also explains why it worked in manual testing: the PNG that downloaded successfully evidently never crossed a host boundary, so the latent failure stayed hidden.

**P3, fixed download timeout vs `max_bytes`.** Fair — `http.Client.Timeout` is wall-clock and includes reading the body, so one value either kills a legitimate 10 MB transfer on a slow link or lets a 1 MB one hang. Now scales from a 30s base at a pessimistic 128 KB/s floor, capped at 5 minutes.

**P3, "stream the response to disk".** Also correct, and I missed this one in the first round. The action has no streaming path; the content is base64-encoded into the response body. Both messages now say to raise `max_bytes` and, separately, to save and decode the response rather than read it into a context.

PR #654, opened 2026-08-07
fix(adapters): make downloads byte-exact; add gmail raw message

ericlevine:
Follow-up to #653, which established byte-exact downloads for Slack. This applies the same rule to the adapters that predate it, and adds a raw-message endpoint for Gmail.

## Byte-exact downloads

Downloads returned text *unencoded* when the content type looked textual. That cannot round-trip a file:

- `format.SanitizeText` **strips HTML** — an `.html` download returned the prose with the markup deleted — and truncates at 200k runes.
- Even unsanitized, bytes in a JSON string lose anything not valid UTF-8 to U+FFFD, silently corrupting legacy-encoded CSVs and every binary format.

Both failures are success-shaped: the caller gets a `Downloaded foo.html` summary and altered content.

| Path | Change |
|---|---|
| `dropbox download_file` | text branch removed |
| `onedrive download_file` | text branch removed |
| `drive export_file` | **was sanitizing exported bytes** — export targets include PDF and the OOXML formats, which `SanitizeText` destroys outright |
| `drive` export sheet tab | was stripping markup out of cell values and truncating the sheet |

`drive get_file` is deliberately **untouched**: its `content` is an explicit preview for text mimetypes on a metadata call, not a download. Changing it would alter a field that is meant to be a lossy summary.

## Gmail `get_message_raw`

An **alternative** endpoint, not a replacement. `get_message` returns a readable projection (from/subject/body/labels), which is what most callers want and is far cheaper — it stays the right default. The raw endpoint serves what a projection cannot:

- header and DKIM verification
- verbatim re-send or archival
- MIME parts the projection drops

Gmail emits URL-safe base64; this re-encodes as standard base64 so it matches the `encoding: base64` contract every other content action uses. Bounded by `max_bytes` (1 MB default, 10 MB ceiling) because a raw message carries its attachments inline and can be much larger than the parsed form — with a bounded read rather than the unbounded `io.ReadAll` in the shared `gmailGET`.

## Testing

Full suite green. New Gmail tests assert the raw bytes round-trip exactly — including a DKIM header and `<b>hi</b>` markup that `get_message` would drop or strip, plus a non-UTF-8 byte — and cover padded/unpadded URL-safe input, the size guard, and the error paths. The Drive sheet-export test now decodes before asserting.

**Contract change:** callers of `dropbox download_file`, `onedrive download_file`, and `drive export_file` that previously received plain text for textual files now receive base64. This is the same break #653 made for Slack, applied for the same reason.

**Not verified against live APIs** — all tests are stubbed transports. Worth confirming a Drive→PDF export and a Gmail raw fetch with an attachment in staging.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/654?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/adapters/definitions/dropbox.yaml
+++ b/internal/adapters/definitions/dropbox.yaml
@@ -75,6 +75,8 @@ actions:
     display_name: "Download file"
     risk: { category: read, sensitivity: low, description: "Download a file's content from Dropbox" }
     override: go
+    # Content is returned base64-encoded for every file type, so the file
+    # round-trips byte for byte. Decode it before use.
     params:
       path:
         type: string
--- a/internal/adapters/definitions/dropbox.yaml
+++ b/internal/adapters/definitions/dropbox.yaml
@@ -82,6 +82,9 @@ actions:
         type: string
         required: true
         description: "File path in Dropbox (e.g. /Documents/report.pdf)"
+      max_bytes:
+        type: int
+        description: "Maximum bytes to download (default 1 MB, max 10 MB). Oversized files are refused, not truncated."
 
   upload_file:
     display_name: "Upload file"
--- a/internal/adapters/definitions/google_drive.yaml
+++ b/internal/adapters/definitions/google_drive.yaml
@@ -67,6 +67,9 @@ actions:
     override: go
     params:
       file_id: { type: string, required: true }
+      max_bytes:
+        type: int
+        description: "Maximum bytes to download (default 1 MB, max 10 MB). Oversized files are refused, not truncated."
 
   create_file:
     display_name: "Create file"
@@ -102,6 +105,9 @@ actions:
       sheet_name:
         type: string
         description: "Google Sheets only: name of the tab to export when mime_type is text/csv or text/tab-separated-values. Without this, only the first tab is returned."
+      max_bytes:
+        type: int
+        description: "Maximum bytes to download (default 1 MB, max 10 MB). Oversized files are refused, not truncated."
 
   search_files:
     display_name: "Search files"
--- a/internal/adapters/definitions/google_gmail.yaml
+++ b/internal/adapters/definitions/google_gmail.yaml
@@ -44,6 +44,27 @@ actions:
     params:
       message_id: { type: string, required: true }
 
+  # Alternative to get_message, not a replacement. get_message returns a
+  # readable projection (from/subject/body/labels) and is what most callers
+  # want. This returns the original RFC 5322 bytes, base64-encoded, for the
+  # cases a projection cannot serve: header or DKIM verification, verbatim
+  # re-send and archival, or MIME parts the projection drops.
+  get_message_raw:
+    display_name: "Get raw message"
+    risk: { category: read, sensitivity: low, description: "Read a specific email message in its original RFC 5322 form" }
+    scopes: ["https://www.googleapis.com/auth/gmail.readonly"]
+    override: go
+    params:
+      message_id: { type: string, required: true }
+      max_bytes:
+        type: int
+        description: >
+          Maximum decoded message size (default 1 MB, max 10 MB). A raw
+          message carries its attachments inline, so it can be much larger
+          than the parsed form. Content is returned base64-encoded, which
+          inflates it by ~4/3 — only raise this when saving the response to a
+          file rather than reading it into a context.
+
   get_thread:
     display_name: "Get thread"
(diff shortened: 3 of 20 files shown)


ericlevine:
All ten addressed in 0616f020. The three P1s were correct and the same defect, and it was worse than "still truncated" — I'd made it *look* fixed.

**P1 × 3 — downloads still capped at `MaxBodyLen`.** I changed the encoding but left every read at `io.LimitReader(resp.Body, format.MaxBodyLen)`. So a 300 KB file was still truncated at 200 KB, and base64-encoding the prefix made it decode cleanly while `size` reported the original from metadata. Nothing downstream could detect it. That's precisely the silent-corruption class this PR exists to remove, one layer below where I fixed it.

Downloads now bound by `max_bytes` (1 MB default, 10 MB ceiling) and **refuse** overflow rather than returning a prefix, matching Slack. Read errors are propagated too, so a connection dropped mid-transfer isn't reported as a short success.

Note `drive download_file` had the identical cap and wasn't flagged — fixed as well. OneDrive needed it on both the pre-signed and `/content` paths.

**P2 — unbounded sheet-tab export.** Correct, and it had no limit at all rather than a wrong one. Bounded before encoding.

**P2 — fractional `max_bytes`.** Correct: `1.9` silently became `1`, and `0.5` became `0` and then failed as "must be positive", which is a misleading error for a bad input. Now rejected explicitly.

**P3 — duplicated `resolveMaxBytes`.** Good call, hoisted to `format.ResolveMaxBytes` with `format.ReadBounded` alongside it. One place now owns the default, the ceiling, and the validation, so Slack and Gmail can't drift. Both duplicated tests collapsed into one table-driven test in `format`.

**P3 — Gmail envelope bound trips before the decoded-size check.** Right; a message well over the limit produced "response exceeds N bytes" with no path forward. It now returns the same actionable `max_bytes` guidance.

**P3 × 2 — dead `isTextContent`.** Removed from both packages.

**P2 — `max_bytes` not exposed to consumers.** Partly deferred, deliberately. I added the parameter to each definition so the limit is reachable and documented, but you're right that `description:` on a param is *silently dropped*: `yamldef.Param` has no `Description` field at all, so `yaml.v3` discards it. That affects every `description:` on every param across all definitions, not just this one — Dropbox and Drive already write them. Surfacing them means adding the field, threading it through `ActionMeta`, and emitting it in the catalog. That's a systemic fix worth its own PR rather than being buried in this one; happy to open it next if you want.

Added a regression test asserting a 300 KB download now returns intact and that exceeding `max_bytes` errors instead of truncating.

ericlevine:
All four addressed in 56cbced8. The OneDrive one was a genuine defect I introduced.

**P2 — misplaced `max_bytes`.** Correct and precisely diagnosed. A scripted edit matched the wrong insertion point and dropped the block into `search_files`' `response:` block instead of `download_file`'s `params:`. Exactly as described, the failure is silent: `ResolveMaxBytes` falls back to the 1 MB default and no caller can opt into a larger download.

Moved to the right action. More usefully, `TestDownloadActionsExposeMaxBytes` now asserts the **exact** set of actions carrying `max_bytes` across all loaded definitions — both that the five download actions have it and that nothing else does. It caught `drive export_file` as a legitimate extra while I was writing it, which is the behaviour I wanted. A misplaced block now fails CI rather than shipping silently.

**P2 — sheet export memory.** Right that bounding only the serialized CSV was too late: the values JSON is fully materialized first, so the memory is spent before the check runs. Added `apiGETLimited` and bounded the values response itself, with headroom since the JSON envelope exceeds the CSV it produces.

**P3 — out-of-range float.** Correct on the language point: float-to-int64 conversion outside range is implementation-defined, and `1e19` only *happened* to land negative and get caught by the positivity check. Now range-checked in float space before converting.

**P3 — impossible recovery advice at the ceiling.** Good catch. Added `format.OverflowMessage`, which suggests raising `max_bytes` only when headroom exists and otherwise reports the hard limit. Applied across Slack, Dropbox, Drive, OneDrive, and Gmail so the wording can't diverge — previously each had its own copy of the string, which is how the ceiling case slipped through in the first place.

PR #657, opened 2026-08-09
Request packs, auto-refill, and annual billing

ericlevine:
## What

Adds the UI for prepaid request packs and auto-refill, makes the annual plan purchasable, and introduces a hook so cloud layers can gate task creation on every path rather than just the HTTP route.

The API already supported packs — nothing rendered them, so a customer refused with `QUOTA_EXHAUSTED` had no way to buy more.

## Task creation hook

`ServerOptions.TaskHooks.BeforeCreate` runs before any task is created. The three creation paths don't share a route — `POST /api/tasks`, MCP tool dispatch, and inline creation by the LLM proxy inside `POST /api/v1/messages` — so gating at the router only covered the first and left the other two as silent bypasses. All three now call one `guardTaskCreate` chokepoint.

Nil (the open-source default) never refuses, so behaviour here is unchanged.

## Billing page

- The usage bar measures the monthly allowance only, so a full bar always means the same thing. Pack balance sits below it as a separate reserve — it's bought rather than granted and never resets with the month — with an inline buy action.
- Auto-refill distinguishes "no card saved" from "switched off", and warns when it's on but can't charge. The toggle stays usable without a card so a customer whose card lapsed can still switch it off.
- Billing controls key off `billed_plan` rather than the entitlement. A grandfathered account that also pays for Pro previously lost "Manage subscription" entirely, leaving no way to update a card or cancel.

## Pricing page

- Monthly/annual toggle. The annual price existed in the API and in Stripe, but no UI path could reach it — and the card advertised a rate the checkout button didn't charge.
- `isCurrent` compares against `billed_plan`, so an existing subscriber is no longer shown an enabled "Get Pro" that would create a second subscription.

## Copy

Removes "Uncapped during early access" from the signup path and the stale $199 Pro price. Both became false once the limit was enforced.

## Notes

`lib/quota.ts` is a single source of truth for how close an account is to stopping, so the global banner and the billing page can't disagree about who is out of requests.

Verified in a browser against a running server: the pricing toggle switches $120/$150 correctly, the pack grid and auto-refill states render from live API data, and the blocked banner appears app-wide for an exhausted account.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/657?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/api/handlers/task_quota_hook_test.go
+++ b/internal/api/handlers/task_quota_hook_test.go
@@ -1,0 +1,43 @@
+package handlers
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/clawvisor/clawvisor/pkg/store"
+)
+
+// The hook must apply on EVERY creation path. Gating at the HTTP router only
+// covered one of three: MCP tool dispatch and inline LLM-proxy creation never
+// touch a task route, so an out-of-quota account kept minting tasks through
+// them. guardTaskCreate is the single chokepoint all three now share.
+func TestGuardTaskCreate(t *testing.T) {
+	h := &TasksHandler{}
+	agent := &store.Agent{ID: "a1", UserID: "u1"}
+
+	// No hook installed (open-source build) — never refuses.
+	if err := h.guardTaskCreate(context.Background(), agent); err != nil {
+		t.Fatalf("no hook should allow, got %v", err)
+	}
+
+	blocked := errors.New("out of requests")
+	var sawUser string
+	h.SetBeforeTaskCreate(func(_ context.Context, a *store.Agent) error {
+		sawUser = a.UserID
+		return blocked
+	})
+
+	if err := h.guardTaskCreate(context.Background(), agent); !errors.Is(err, blocked) {
+		t.Fatalf("hook error must propagate, got %v", err)
+	}
+	if sawUser != "u1" {
+		t.Errorf("hook saw user %q, want u1 — quota is charged to the agent's owner", sawUser)
+	}
+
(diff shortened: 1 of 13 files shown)


ericlevine:
Thanks — went through all three.

**Fixed: interval sent for plans that don't offer it** (`11d60f9`). Correct catch, and it's the same advertised-price-vs-charged-price mismatch this PR exists to remove. The cadence is now sent only when the plan actually has that price option; otherwise the server applies its default. Verified against the real catalog shape: Pro sends the selected cadence, Free and Enterprise send none.

**Not doing: runtime-event promotion** (`RuntimeHandler.createTaskFromRuntimeEvent`) and the other direct `CreateTask` callers in `runtime.go` / `runtime_controls.go`. Deliberate scope call — the runtime-proxy paths are out of scope for this change. The finding is accurate though, so worth noting plainly: **the guard is not universal today.** It covers `POST /api/tasks`, MCP tool dispatch, and the inline LLM-proxy entries; runtime-event promotion is not gated.

**Not doing: the plain `CreateInlineApprovedTask` variant**, same reason — it's on the LLM-proxy path. Also accurate: it delegates to the shared `createInlineApprovedTask` without the guard its sibling has, so the two can diverge.

I've toned down the "every task-creation path" wording in the hook's doc comment rather than leave a claim the code doesn't support.

PR #662, opened 2026-08-11
fix(adapters): stop discarding param descriptions

ericlevine:
`yamldef.Param` had no `Description` field, so `yaml.v3` silently dropped every `description:` written under a parameter.

Definitions carry **46 of them across 8 services** — all documenting nothing:

| Service | Dropped |
|---|---|
| dropbox | 12 |
| microsoft.outlook | 9 |
| google.gmail | 8 |
| google.sheets | 5 |
| microsoft.onedrive | 4 |
| google.calendar | 4 |
| slack | 2 |
| google.drive | 2 |

The text was written, reviewed, and maintained for as long as the schema has existed, and never reached a caller. Nothing failed — `yaml.v3` ignores unknown keys by default, so the only symptom was an agent reading a catalog entry thinner than the YAML suggested.

Found while adding a `max_bytes` param whose description explains a non-obvious limit (1 MB default / 10 MB ceiling). Without this, callers see `max_bytes?` and its type but no way to learn the bounds short of triggering the error.

## Change

Adds the field and threads it through both metadata carriers:

- `adapters.ParamInfo` — pre-execution validation
- `adapters.ParamMeta` — the service catalog

then renders it in the catalog's per-parameter detail, matching what the local-services path (`LocalCatalogParam`) already did. The two renderers sat side by side in `skill.go`, one printing descriptions and one silently unable to.

Descriptions are flattened to one line first: YAML block scalars (`>` and `|`) preserve newlines, which would break out of the markdown list item they're rendered inside.

## Testing

`TestParamDescriptionsSurviveParsing` walks every embedded definition and asserts each parsed description reaches both carriers — currently 46.

It also **fails if fewer than 20 are found**. Without that floor, a silent parsing regression would leave nothing to iterate over, every assertion would pass vacuously, and the test would stay green while testing nothing. That's the same failure mode this PR fixes, so it seemed worth not reintroducing in the test itself.

`TestFlattenParamDesc` covers folded, literal, and padded inputs.

Full suite green. No behaviour change beyond catalog output; the field is additive and optional.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/api/handlers/skill.go
+++ b/internal/api/handlers/skill.go
@@ -437,7 +437,7 @@ func (h *SkillHandler) writeLocalServiceDetail(buf *strings.Builder, ctx context
 				extras := []string{p.Type, reqTag}
 				buf.WriteString(fmt.Sprintf("- `%s` (%s)\n", p.Name, strings.Join(extras, ", ")))
 				if p.Description != "" {
-					buf.WriteString(fmt.Sprintf("  %s\n", p.Description))
+					buf.WriteString(fmt.Sprintf("  %s\n", flattenParamDesc(p.Description)))
 				}
 			}
 		}
@@ -507,6 +507,9 @@ func (h *SkillHandler) writeServiceDetail(buf *strings.Builder, ctx context.Cont
 							extras = append(extras, fmt.Sprintf("max: %d", *p.Max))
 						}
 						buf.WriteString(fmt.Sprintf("- `%s` (%s)\n", p.Name, strings.Join(extras, ", ")))
+						if p.Description != "" {
+							buf.WriteString(fmt.Sprintf("  %s\n", flattenParamDesc(p.Description)))
+						}
 					}
 				}
 				buf.WriteString("\n")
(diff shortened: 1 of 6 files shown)


ericlevine:
Both addressed in 207821a1.

**P3 — the two renderers were not actually in sync.** Correct, and the comment was the wrong half to trust: `writeLocalServiceDetail` printed `p.Description` raw, so it carried the same latent newline bug the adapter path was being fixed for. Flattened there too, and deleted the comment rather than leave a claim that can drift out of date again.

**P3 — the `described < 20` floor.** Also correct, and it's the more interesting one: a floor is exactly the failure mode this PR exists to fix. If one service lost its descriptions, those params never enter the loop, so every assertion passes vacuously and the test goes green having checked nothing.

Rather than pin an exact count (46 today, brittle against any definition edit), the test now parses each definition **generically** — bypassing `yamldef` entirely — counts param descriptions per service straight from the YAML, and requires the typed schema to have captured exactly that. Ground truth comes from the file, so adding or removing a description needs no test edit, while dropping one on the floor fails.

Verified by simulating the real regression: renaming the schema's yaml tag so parsing drops the field. The test reports

```
param descriptions captured per service = map[], want map[dropbox:12 google.calendar:4
google.drive:2 google.gmail:8 google.sheets:5 microsoft.onedrive:4 microsoft.outlook:9 slack:2]
```

My first attempt at that check was wrong, worth noting: I deleted descriptions from a YAML file and the test passed — correctly, since that lowers both sides. Removing a description is a legitimate edit; the schema silently discarding one is not, and that's what's now covered.

PR #680, opened 2026-09-04
fix(notify): make decision delivery at-least-once with an explicit ack

ericlevine:
Approvals were lost whenever a Cloud Run instance stopped. `BRPOP` is a destructive read, so between taking a decision and finishing with it the only copy lived in a goroutine — a scale-down or rolling deploy dropped a human's decision with no trace. On staging this lost **roughly half of all approvals**: the prompt stayed actionable while the task resolved.

Not Slack-specific. The same bus drives Telegram, where a message that never gets edited is far less visible than buttons that stay live.

## Why shutdown ordering could not fix it

Three attempts in #679 each moved the loss rather than closing it:

- a **requeue** on cancellation — races process exit, so the write may not land
- a **drain** for the subscriber — returns after a decision is handed to a consumer that is itself exiting, which then fails with `context canceled`
- an **inflight WaitGroup** — its `Add` races the very `Wait` meant to bound it

None can work. There is no instant at which the decision is both off the queue and durably owned, so no sequencing makes that interval safe. The missing property was an acknowledgement boundary.

## What changed

`BLMOVE` takes a payload from the queue onto a processing list in one atomic step, so it is never only in memory. The consumer acks once handled, which retires it. Anything left unacked past a visibility timeout is returned to the queue by a reclaimer, which also runs at startup to recover what a previous instance left behind.

Shutdown ordering stops mattering: an instance that dies mid-handler simply does not ack, and `requeue`, `Drain` and the WaitGroup are deleted rather than fixed.

## At-least-once

Handlers must be idempotent. The approval paths already are — they CAS from `pending_approval`, so a duplicate loses the race and no-ops.

Acking is deliberate on both paths. A **handler error acks**: a decision that was delivered and rejected (an expired approval, a lost CAS) is not a lost one, and redelivering it would loop forever. Only an **unfinished** handler goes unacked, which is exactly the case that was being lost.

## Interface change

`DecisionBus.Subscribe` now returns `<-chan Delivery` rather than bare decisions, so the ack can travel with the item. `LocalDecisionBus` acks are no-ops — with one process there is nothing to redeliver from — and its channel is now unbuffered for the same reason the Redis one is.

## Tests

Three cover the properties that matter: an unacked decision is redelivered rather than lost; an acked one is not; and one still inside its visibility timeout is left alone rather than run twice.

<a href="https://cubic.dev/pr/clawvisor/clawvisor/pull/680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -1884,17 +1884,22 @@ func (s *Server) consumeNotifierDecisions(ctx context.Context, ch <-chan notify.
 				}
 			}
 			if err != nil {
-				s.logger.WarnContext(ctx, "notifier decision failed",
+				// Deliberately not acked. A handler error does not
+				// distinguish "delivered and rejected" — an expired
+				// approval, a lost CAS — from "transiently failed": a
+				// database outage, or this context being cancelled
+				// mid-handle during a rolling deploy. Acking here would
+				// discard the second kind, which is the loss this bus
+				// exists to prevent. Redelivery is bounded by
+				// maxDeliveryAttempts, so a decision that genuinely
+				// cannot be handled is dropped loudly rather than
+				// retried forever.
+				s.logger.WarnContext(ctx, "notifier decision failed, will be redelivered",
 					"type", d.Type, "action", d.Action,
 					"target_id", d.TargetID, "err", err)
+				continue
 			}
 
-			// Acked either way. A handler error is a decision that was
-			// delivered and rejected — an expired approval, a lost CAS —
-			// not one that went missing, and redelivering it would just
-			// fail again on a loop. Only an unfinished handler should be
-			// retried, and that is the case where this line is never
-			// reached.
 			delivery.Ack()
 
 		}
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -1833,15 +1833,16 @@ func (s *Server) handleClawvisorKeys(w http.ResponseWriter, r *http.Request) {
 
 // consumeNotifierDecisions reads from the notifier's decision channel
 // and routes approve/deny decisions to the appropriate handler.
-func (s *Server) consumeNotifierDecisions(ctx context.Context, ch <-chan notify.CallbackDecision) {
+func (s *Server) consumeNotifierDecisions(ctx context.Context, ch <-chan notify.Delivery) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case d, ok := <-ch:
+		case delivery, ok := <-ch:
 			if !ok {
 				return
 			}
+			d := delivery.Decision
 			// Logged on arrival: if a decision never reaches here, the
 			// prompt is resolved by some other route and no amount of
 			// instrumentation further down will show why.
@@ -1887,6 +1888,14 @@ func (s *Server) consumeNotifierDecisions(ctx context.Context, ch <-chan notify.
 					"type", d.Type, "action", d.Action,
 					"target_id", d.TargetID, "err", err)
 			}
+
+			// Acked either way. A handler error is a decision that was
+			// delivered and rejected — an expired approval, a lost CAS —
+			// not one that went missing, and redelivering it would just
+			// fail again on a loop. Only an unfinished handler should be
+			// retried, and that is the case where this line is never
+			// reached.
+			delivery.Ack()
 
 		}
 	}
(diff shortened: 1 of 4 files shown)


ericlevine:
All four findings are valid, and all four are mine. Fixed in the latest push.

They share one root cause I should have caught while writing this: **the raw JSON payload was the delivery identity**, and the operations around it were not atomic.

**P0 — unmarked processing entries.** `BLMOVE` and the `ZADD` that follows are two commands, and my comment claiming a failure there "only costs a slower reclaim, never the decision" was simply wrong — I asserted the safety property I most needed to check. Rather than making the two commands atomic, `reclaim` now adopts processing entries that have no marker (`ZADDNX`, so a real start time recorded concurrently is never clobbered) and ages them out normally. That turns the window from lossy into merely slow.

**P1 — duplicate payloads.** Each publish now carries a UUID in an envelope, so two publishes of the same decision are two distinct entries with two markers.

**P2 — non-atomic requeue.** `LREM`/`LPUSH`/`ZREM` are now one Lua script with the push gated on `LREM > 0`, so an ack and a reclaim racing are mutually exclusive — whichever lands first wins and the loser is a no-op. This also covers the new race the P0 fix introduces, where a marker can be adopted for an entry that was acked mid-scan.

**P1 — acking on handler error.** You are right and I was wrong. I flagged this as the line to push on, and the distinction you drew is the one I missed: a handler error does not separate "delivered and rejected" from "transiently failed". A DB outage or a context cancelled mid-handle during a rolling deploy is exactly the loss this PR exists to prevent, and I re-introduced it one layer up. Errors no longer ack. The loop-forever risk that motivated my original choice is handled with a five-attempt cap carried in the envelope — an unhandleable decision is dropped with an ERROR log rather than retried indefinitely.

Each fix has a regression test, and I verified by reverting each fix individually that its test fails:

| revert | failing test |
|---|---|
| drop `adoptUnmarked` | `UnmarkedInFlightDecisionIsRecovered` |
| constant instead of per-publish ID | `IdenticalDecisionsDoNotCollide` |
| ungate the `LPUSH` | `StaleMarkerDoesNotResurrectAnAckedDecision` |
| disable the attempt cap | `GivesUpAfterRepeatedRedelivery` |

One gap worth naming: the `consumeNotifierDecisions` change is untested. It is a three-line control-flow change and the package has no existing harness for that function, so I left it to review by eye rather than building one — happy to add it if you would rather have the coverage.

ericlevine:
Both valid, both fixed.

**P2 — legacy flat payloads.** This is the more serious of the two and I should have caught it: the format change was not rollout safe, and a rolling deploy is guaranteed to run both builds against one queue. A flat decision from the previous build unmarshalled into the nested envelope *cleanly but emptily* — no decision type, so the consumer matched no case, `err` stayed nil, and it was acked away. A silent loss of exactly the kind this PR exists to prevent, triggered by the deploy that ships the fix.

Rather than special-casing the decode, the envelope now **embeds** the decision instead of nesting it, which fixes both directions rather than only the one you named:

- an **old** instance reading a **new** payload finds every decision field at the top level and ignores the underscored `_delivery_id` / `_attempts`
- a **new** instance reading an **old** payload fills the embedded struct and simply has no delivery ID

The reverse direction would have lost decisions just as silently, and a decode-side fix alone would not have covered it.

I also stopped delivering payloads that decode but name no decision. Handing one to the consumer means it matches no case, is treated as handled, and gets acked out of existence — so those are set aside instead.

**P1 — deleting exhausted deliveries.** You are right that this is not a call the bus can make. Handler errors cannot distinguish a permanently unhandleable decision from a fifteen-minute database outage, so dropping after five cycles was me resolving that ambiguity by destroying evidence. Exhausted deliveries and unparseable payloads now move to a bounded dead-letter list (`clawvisor:decisions:dead`, capped at 1000 via `LTRIM`) in the same atomic step as the removal. Nothing is deleted; retiring one is destroying something a human decided.

Mutation-tested as before — each fix reverted individually, confirming its test catches it:

| revert | failing test |
|---|---|
| nest the decision under a `decision` key | `LegacyFlatPayloadIsStillDelivered` **and** `PublishedPayloadIsReadableByThePreviousFormat` |
| delete on exhaustion instead of dead-lettering | `GivesUpAfterRepeatedRedelivery` |
| deliver payloads with no decision type | `UnrecognisedPayloadIsDeadLetteredNotAcked` |

Reverting to the nested envelope fails both rollout tests with an empty decision — the exact scenario you described.

